PR #52, opened 2022-09-07
Add platform support for upcoming RP2040-based ZuluSCSI

morio:
Adding RP2040 code to public repo


aperezbios:
I've tested the CI-built firmware on an RP2040 board and it works fine. Merging.

PR #78, opened 2022-10-04
Remove apple quirks from defaulting to right align

morio:
Right-aligned strings break the ability of Apple's Disk Utility to initialize drives. Reverting this behavior.


PetteriAimonen:
Based on code review, I think this commit does the opposite of what is desired.
The `img.rightAlignStrings` defaults to `-1` which was used to indicate "no value set by user".
Now that the `if (rightAlign < 0)` is removed, I think `formatDriveInfoField()` will interpret `-1` as `true`.

aperezbios:
Agreed. We overlooked this. Fixed in 9d7571

PR #118, opened 2022-12-07
GPT partition support

PetteriAimonen:
Especially Mac OS X Disk Utility formats SD cards as GPT partition table by default.
This pull request adds support in SdFat to access these SD cards also.

Also upgraded to SdFat 2.2.0 in all configurations.
Upstream pull-request for GPT support in SdFat is here: https://github.com/greiman/SdFat/pull/406
For now using this fork for ZuluSCSI: https://github.com/rabbitholecomputing/SdFat


PetteriAimonen:
Hmm, I guess better not merge this yet - it seems the GPT support is not going to be accepted upstream in SdFat, so I'll have to rework this to avoid future maintenance problems.

Issue #124, opened 2022-12-20
"Failed to load image" when file is marked read-only


PetteriAimonen:
When image file is marked read-only, ZuluSCSI will refuse to open it and report "Failed to load image" in log.

At least there should be a better error message for this.

Ideally, read-only images should be opened in read-only mode and shown as read-only drive to the SCSI host.

morio:
I'll see if I can get it to report a better error message

morio:
Pull request merged

PR #144, opened 2023-02-15
Open image as read-only if attribute set (#124)

morio:
A disk image will be opened as read-only if the read-only attribute is set on the image file. If the SCSI disk is been written to a read-only image, the "Write protected" sense code will be returned. If that is somehow bypassed the ImageBackingStore write method will log an error and return 0 bytes written.

Issue #124

### DIFF
--- a/src/ZuluSCSI_disk.cpp
+++ b/src/ZuluSCSI_disk.cpp
@@ -272,6 +272,7 @@ public:
     {
         m_israw = false;
         m_isrom = false;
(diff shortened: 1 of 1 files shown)


morio:
Hi @PetteriAimonen, could you give a quick code review of this pull request?

aperezbios:
Tested on a couple of different platforms, and it seems to be working as expected.

PetteriAimonen:
Looks good to me.

PR #147, opened 2023-02-22
RP2040: Add .ini option to enable/disable parity check.

PetteriAimonen:
Parity bit will always be transmitted to the host. RP2040 platform can optionally check received parity. Some hosts do not implement parity, so this option allows to disable the check:

    [SCSI]
    EnableParity = 0

The log file will indicate either `-- EnableParity is on` (default) or `-- EnableParity is off` on platforms that support the check.

**Note:** I haven't yet tested this on real hardware myself. Please download test build from here and check before merging:
https://github.com/ZuluSCSI/ZuluSCSI-firmware/actions/runs/4240160312

### DIFF
--- a/lib/ZuluSCSI_platform_RP2040/ZuluSCSI_platform.h
+++ b/lib/ZuluSCSI_platform_RP2040/ZuluSCSI_platform.h
@@ -21,6 +21,7 @@ extern const char *g_azplatform_name;
 #define PLATFORM_OPTIMAL_LAST_SD_WRITE_SIZE 8192
 #define SD_USE_SDIO 1
 #define PLATFORM_HAS_INITIATOR_MODE 1
+#define PLATFORM_HAS_PARITY_CHECK 1
 
 // NOTE: The driver supports synchronous speeds higher than 10MB/s, but this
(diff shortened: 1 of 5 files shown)


the-claw:
Tested on RP2040 h/w using a SCSI-1 (SASI) host with no parity support (Sharp X68000 Ace).

When EnableParity is 0 in zuluscis.ini the firmware works as expected, reads and writes can occur without error.

When EnableParity is 1 or omitted from the config, the host receives a parity error (see [log file](https://github.com/ZuluSCSI/ZuluSCSI-firmware/files/10834034/zululog.txt) ) on any command. This behaviour differs from the previous fw I tested (1.2.0), which ignored parity errors on all reads and only reported parity error on write. However, I think the change in behaviour makes sense because if parity is enabled it should be enforced for all commands.

aperezbios:
@the-claw thank you for testing and reporting the results. We really appreciate it.

PetteriAimonen:
@the-claw Yeah, previously it wasn't consistent, because the SCSI command level code wasn't given the flag to check parity, so for some commands it disregarded the parity check status.

PR #159, opened 2023-03-18
Boot delay improvement

PetteriAimonen:
Related to #138, reduce the boot delay from about 1 second to less than 0.3 seconds.
This allows using "SCSI Timeout: short" setting on Amiga.


aperezbios:
I'll test this more post-merge

PR #163, opened 2023-03-24
Temp fix for an issue that needs further study

morio:
Certain version strings seem to break the booting of the device.


PetteriAimonen:
Very weird. The custom RP2040 bootloader has been somehow flaky, sometimes it seems to crash soon after writing the flash.

PR #167, opened 2023-03-31
RP2040: Unify platform libraries

PetteriAimonen:
Most of the RP2040 code is shared between the platforms, so unify them into same folder with some #ifdefs.

For each different pinout type, the following files need to be duplicated:

- `ZuluSCSI_platform_gpio_X.h` for the main pinout definitions
- `scsi_accel.pio` if the ACK and REQ pins have different numbers
- `rp2040_sdio.pio` if the SDIO CLK pin has different number

This needs to be tested on the BS2 hardware.


aperezbios:
Thanks, @PetteriAimonen. I'll test this later today on BS2 HW.

aperezbios:
@PetteriAimonen this worked fine on BS2 hardware. I booted it and ran some basic disk benchmarks within MacOS.

PetteriAimonen:
Great :)

PR #177, opened 2023-04-10
Improve git history retention on BS2 platform files

PetteriAimonen:
Rewrote git history to make the BS2 repository commit 8971584485c appear as part of the commit chain.

Does not change code, only git history and some whitespace and comments.


aperezbios:
Thanks, @PetteriAimonen, much appreciated. Looks good.

PR #202, opened 2023-05-09
Improved OMTI support

Kars-de-Jong:
- Command C0 has a non-standard layout, account for this.
 - Implement OMTI-specific request sense.


PetteriAimonen:
I don't have the hardware to test this, but the code looks reasonable and only affects things when `S2S_CFG_QUIRKS_OMTI` is enabled.

Ready to merge in my opinion.

PR #203, opened 2023-05-09
Add rudimentary support for tapes

Kars-de-Jong:
- Add support for reading a single file.
 - Add support for rewinding the tape.
 - Add support for the OMTI 5400 version of the SPACE command.

### DIFF
--- a/src/ZuluSCSI_disk.cpp
+++ b/src/ZuluSCSI_disk.cpp
@@ -610,6 +610,9 @@ struct image_config_t: public S2S_TargetCfg
 
     // Warning about geometry settings
     bool geometrywarningprinted;
+
+    // Tape support
+    uint32_t pos;
 };
 
 static image_config_t g_DiskImages[S2S_MAX_TARGETS];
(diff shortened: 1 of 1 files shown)


PetteriAimonen:
Hi Kars,

There was a merge conflict, as I happened to do a major reorganization of `ZuluSCSI_disk.cpp` at the same time.
It had grown too large, so cdrom support was separated to a different file.

I have adapted your pull request to separate `ZuluSCSI_tape.cpp`.
Could you review #203 and see if it looks good to you?
There is a built binary here: https://github.com/ZuluSCSI/ZuluSCSI-firmware/actions/runs/4945094160

I added a few commands (including write) to make it possible to test it on Linux.
A simple dd to and from the device now works for me.

I made a guess on what "// FIXME - OMTI does not support the FIXED bit." meant by forcing it to use fixed mode with that quirks. Does that look correct?

        bool fixed = scsiDev.cdb[1] & 1;
        if (img.quirks == S2S_CFG_QUIRKS_OMTI)
        {
            fixed = true;
        }

I did not add filemarks support yet, that is probably necessary for storing multiple archives on same tape.
I think it would be most reasonable to have a text file alongside the binary file, listing the filemark positions as numbers.

Kars-de-Jong:
Hi Petteri,

Assuming `fixed = true` for OMTI seems like the correct implementation indeed. I will test the code on the other PR.

Thanks!

aperezbios:
Closing in favor of #203, as it fixed the merge conflicts.

PR #207, opened 2023-05-12
Experimental audio playback support from BIN/CUE files.

saybur:
First attempt at getting audio playback working with the new BIN/CUE support. This has been tested to play audio successfully using 'cdplay' on a modern Linux host. There are bugs and this is not ready for merging, but hopefully the PR can give an idea of potential changes for CD-DA. I'll keep developing this toward something more polished. Known issues:

- Annex C rules for commands to halt audio playback are not supported yet. This prevents playback from being stopped until the initial play command naturally terminates (pausing does work).
- Position tracking is limited to just audio playback, probably needs READ integration.
- Additional mode page support may be needed, I only added a boilerplate 0Eh. Volume control in particular might be pretty useful, but AFAIK there isn't much MODE SELECT support in the firmware?
- Testing has been pretty basic and needs to be expanded, particularly on period-correct systems.
- Testing of non-audio firmware behavior with audio commands needs to be done.

### DIFF
--- a/lib/SCSI2SD/src/firmware/mode.c
+++ b/lib/SCSI2SD/src/firmware/mode.c
@@ -220,6 +220,23 @@ static const uint8_t ControlModePage[] =
 0x00, 0x00 // AEN holdoff period.
 };
 
+#ifdef ENABLE_AUDIO_OUTPUT
+static const uint8_t CDROMAudioControlParametersPage[] =
+{
+0x0E, // page code
+0x0E, // page length
+0x04, // 'Immed' bit set, 'SOTC' bit not set
+0x00, // reserved
+0x00, // reserved
+0x80, // 1 LBAs/sec multip
+0x00, 0x4B, // 75 LBAs/sec
+0x03, 0xFF, // output port 0 active, max volume
+0x03, 0xFF, // output port 1 active, max volume
+0x00, 0x00, // output port 2 inactive
+0x00, 0x00 // output port 3 inactive
+};
+#endif
+
 static const uint8_t SequentialDeviceConfigPage[] =
 {
 0x10, // page code
@@ -495,6 +512,21 @@ static void doModeSense(
 		pageIn(pc, idx, ControlModePage, sizeof(ControlModePage));
 		idx += sizeof(ControlModePage);
 	}
+
+#ifdef ENABLE_AUDIO_OUTPUT
+	if ((scsiDev.compatMode >= COMPAT_SCSI2)
+		&& (scsiDev.target->cfg->deviceType == S2S_CFG_OPTICAL)
+		&& (pageCode == 0x0E || pageCode == 0x3F))
+	{
+		pageFound = 1;
+		pageIn(
+			pc,
+			idx,
+			CDROMAudioControlParametersPage,
+			sizeof(CDROMAudioControlParametersPage));
+		idx += sizeof(CDROMAudioControlParametersPage);
+	}
+#endif
 
 	if ((scsiDev.target->cfg->deviceType == S2S_CFG_SEQUENTIAL) &&
 		(pageCode == 0x10 || pageCode == 0x3F))
--- a/lib/ZuluSCSI_platform_RP2040/audio.h
+++ b/lib/ZuluSCSI_platform_RP2040/audio.h
@@ -32,6 +32,20 @@ extern "C" {
 // these must be divisible by 1024
 #define AUDIO_BUFFER_SIZE 8192 // ~46.44ms
 
+/*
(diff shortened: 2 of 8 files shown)


src/ImageBackingStore.cpp
```diff
@@ -313,3 +313,15 @@ void ImageBackingStore::flush()
         m_fsfile.flush();
     }
 }
+
+size_t ImageBackingStore::name(char* filename, size_t len)
+{
+    if (!m_israw && !m_isrom)
```
PetteriAimonen, line 319: Could be just m_fsfile.isOpen(). The raw mapping also keeps the file open now, because access to 2352 byte sectors may require unaligned access.

lib/SCSI2SD/src/firmware/mode.c
```diff
@@ -220,6 +220,23 @@ static const uint8_t ControlModePage[] =
 0x00, 0x00 // AEN holdoff period.
 };
 
+#ifdef ENABLE_AUDIO_OUTPUT
```
PetteriAimonen, line 223: We could consider adding some hook to `mode.c` to pass mode page requests to ZuluSCSI_cdrom.cpp. Something like `void scsiCDRomRequestModePage(uint8_t pageCode, int *pageFound, int *idx)` that would write to `scsiDev.data`. I would like to keep modifications to SCSI2SD files to minimum, to make it easier to merge changes back and forth when needed.

lib/ZuluSCSI_platform_RP2040/audio.h
```diff
@@ -32,6 +32,20 @@ extern "C" {
 // these must be divisible by 1024
 #define AUDIO_BUFFER_SIZE 8192 // ~46.44ms
 
+/*
+ * Status codes for audio playback, matching the SCSI 'audio status codes'.
+ *
+ * The first two are for a live condition and will be returned repeatedly. The
+ * following two reflect a historical condition and are only returned once.
+ */
+enum audio_status_code {
```
PetteriAimonen, line 41: Would these parts be common to all platforms if another platform gets audio support in future?
If yes, it could be better to make `src/ZuluSCSI_audio.h` for the common parts and interface.

PetteriAimonen (commented):
Looks good overall. Left a few small notes, but nothing critical.
Let us know when you consider it ready for merging.

src/ImageBackingStore.cpp
saybur: Great, I had thought these files were closed when switching between targets. If they're staying open now that'd be easier. Will change.

lib/SCSI2SD/src/firmware/mode.c
saybur: That'd sounds fine. I mainly wanted to see if you had any intention of creating ZuluSCSI_mode.cpp with more comprehensive support for MODE SELECT functionality. I don't know how many devices even bother trying to set stuff that way so it may be more trouble than it's worth. As of now I'm not finding the current setup restrictive for getting basic playback working. Maybe this could be addressed in a separate PR to add additional playback features.

lib/ZuluSCSI_platform_RP2040/audio.h
saybur: Good idea, I'll resolve that.

src/ImageBackingStore.cpp
saybur: This is done. I have it set up in the audio implementation as a pointer to the ImageBackingStore. I'm not well versed enough in C++ to feel confident about this approach; if you'd prefer something different let me know.

saybur:
Should now be minimally functional across platforms. Testing has shown lots of SCSI command variability across audio playback software, so if merged this probably needs a public 'still experimental' disclaimer for any users. Known bugs include:

- Playback progress displays are wrong on at least the stock Mac and Win9x players.
- SCAN is not implemented, which could lead to incorrect behavior when fast-forwarding/rewinding. Software observed so far doesn't actually use this command, it instead just issues a new play command.
- The code has some guesses about how to report positioning. I may not be understanding the spec here. It is starting and stopping where requested so far.

If this seems like it needs refactoring, more time in the oven, etc, let me know.

lib/SCSI2SD/src/firmware/mode.c
PetteriAimonen: Yeah, can be tidied up in the future if we need more modes.

PetteriAimonen:
Looks good to merge in my opinion.

I agree that it should be considered experimental feature, but I guess the related hardware is also still experimental.

PR #209, opened 2023-05-17
CD-ROM: code consolidation and bugfixes

saybur:
This has several proposed tweaks to CD-ROM code, including a fix for a significant bug with sub-channel MSF address reporting I introduced in my last PR.

In addition, this also consolidates the LBA/MSF conversion code to prepare for a possible adjustment to how MSF addresses are handled. I'll open an issue to document that one since it will likely involve more work.

### DIFF
--- a/src/ZuluSCSI_cdrom.cpp
+++ b/src/ZuluSCSI_cdrom.cpp
@@ -492,7 +492,7 @@ static void doReadTOC(bool MSF, uint8_t track, uint16_t allocationLength)
         if (firsttrack < 0) firsttrack = trackinfo->track_number;
         lasttrack = trackinfo->track_number;
 
-        if (track == 0 || track == trackinfo->track_number)
+        if (track <= trackinfo->track_number)
         {
             formatTrackInfo(trackinfo, &trackdata[8 * trackcount], MSF);
             trackcount += 1;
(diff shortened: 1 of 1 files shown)


src/ZuluSCSI_cdrom.cpp
```diff
@@ -1345,9 +1339,10 @@ static void doReadSubchannel(bool time, bool subq, uint8_t parameter, uint8_t tr
             *buf++ = (lba >= trackinfo.data_start) ? 1 : 0; // Index number (0 = pregap)
             if (time)
             {
+                *buf++ = 0;
                 LBA2MSF(lba, buf);
                 dbgmsg("------ ABS M ", *(buf+1), " S ", *(buf+2), " F ", *(buf+3));
```
PetteriAimonen, line 1344: I think this debug message and line 1360 below may have off-by-one error now. Wouldn't it be buf+0..buf+2?

src/ZuluSCSI_cdrom.cpp
```diff
@@ -606,16 +602,13 @@ static void formatRawTrackInfo(const CUETrackInfo *track, uint8_t *dest)
...
-    if (track->pregap_start > 0)
-    {
-        LBA2MSFRaw(track->pregap_start, &dest[4]);
-    }
-    else
-    {
-        LBA2MSFRaw(track->data_start, &dest[4]);
-    }
-
+    // Next three are ATIME. The spec doesn't directly address how these
```
PetteriAimonen, line 605: Interesting, I was also confused by the spec not telling what to put here. I agree that imitating existing CDs makes sense.

src/ZuluSCSI_cdrom.cpp
```diff
@@ -492,23 +514,20 @@ static void doReadTOC(bool MSF, uint8_t track, uint16_t allocationLength)
         if (firsttrack < 0) firsttrack = trackinfo->track_number;
         lasttrack = trackinfo->track_number;
 
-        if (track == 0 || track == trackinfo->track_number)
+        if (track <= trackinfo->track_number)
```
PetteriAimonen, line 517: Ah, indeed a bug. I missed that track specifies starting track for reporting and not the only track for reporting.

PetteriAimonen (changes requested):
I think there is off-by-one error in debug messages. When that is fixed, I consider this ready to merge.

src/ZuluSCSI_cdrom.cpp
saybur: Ah, definitely. Fixed, thanks for the catch!

PetteriAimonen (approved):

Issue #135, opened 2023-01-31
Physical eject button for changing between CD-ROM iso images


csulaguy:
I suppose this is more of a suggestion than an issue, but is it possible in a future revision to have a hardware pushbutton for the purpose of switching cd-rom images, and have that button tied to a specific SCSI ID in the .ini file? I know ZuluSCSI and RP2040 support multiple images, but not all OSes provide an easy way to eject a drive, since a physical drive would obviously have an eject button on it typically.

PetteriAimonen:
On most models there is an expansion pinheader, so this could even be added to current revisions.

csulaguy:
I presume this would be J303. How hard is it to implement this with a firmware update, so I can add this functionality? Or even the option of using that with more than one SCSI ID (if I'm emulating two or three CD-ROM drives in total)?

aperezbios:
@csulaguy it gets much more complicated to implement for more than one removable/CD-ROM drive, because there's no real way of knowing which SCSI ID it's intended for.

csulaguy:
I know this is an older thread, but how possible will it be to add a momentary switch to some pins on the expansion header, so I can achieve eject functionality with CD-ROMs on a given SCSI ID?

aperezbios:
@csulaguy which ZuluSCSI model do you have? The full size V1.1, or ZuluSCSI RP2040?

csulaguy:
> @csulaguy which ZuluSCSI model do you have? The full size V1.1, or ZuluSCSI RP2040?

Both.

PetteriAimonen:
My suggested minimal implementation would be:

1. Support for two buttons, first one connected between SDA pin and GND, second one between SCL and GND. I think these pins are available on all current models.
2. INI file configuration like:

```
[SCSI2]
EjectButton = 1

[SCSI5]
EjectButton = 2
```

This would allow future configuration for other button functions if we invent any, or use of actual I2C devices because the buttons disturb the bus only when pressed.

csulaguy:
Where would I find SDA/SCL and GND? On 1.1, would it be part of J305 on v1.1? And where on RP2040?

PetteriAimonen:
On full size RP2040, the pins are labeled SDA and SCL on the bottom side of the PCB.
On RP2040 compact, it is on J305 but unlabeled.

[...]

PR #217, opened 2023-05-20
Support for platform eject buttons.

saybur:
Here's a potential option for closing #135. This creates a new platform call that returns the status of any buttons installed on the hardware. Configuration is done via `EjectButton = 1` type options described in the issue.

This also modifies the audio RP2040 variant to treat the unused 'SDA' pin as a pulled-up GPIO to get button state.

Testing on this has been pretty minimal, but it does appear to work properly on my audio test board.


PetteriAimonen (approved):
Looks like a good base to build on and ready to merge.
To be actually useful for most users we'll need to add platform-specific implementations.

The I2C pins can be used to detect button presses when no I2C communication is ongoing (and currently there are no I2C attachments implemented).

PR #229, opened 2023-05-31
CD-ROM image listing from directories

### LINKED ISSUE #218: Feature request: reduced INI editing when adding CD images
The current image system requires either renaming CD images to match the auto-scan format (CD3.bin/CD3.cue), _or_ manually typing the name in the INI file. It would be nice if this could be supplemented by scanning a directory on startup. For example:

```
[SCSI3]
Directory = "/cds"
```

Or even better, if a folder named "CD3" is present, a CD-ROM drive gets auto-configured as if each file within the folder had been listed in the INI file using the `IMG` syntax.

Ideally this would use something intuitive to determine the first image mounted: alphabetical ordering would be great if that doesn't add too much complexity.

saybur:
Proposal to resolve #218 by adding an .ini file option called 'ImgDir' to read images out of a directory at runtime. This implementation follows the approach @PetteriAimonen outlined in the associated issue. It should also give a structure for supporting ini-free configuration in the future.

Basic image iteration works, including ejection events. However, there are  parts (like directory naming) that are incomplete and testing has not yet been completed on different operating systems. If this approach seems workable, I'll continue to develop this and mark it as ready for merging once issues are fixed.

### DIFF
--- a/src/ZuluSCSI_disk.cpp
+++ b/src/ZuluSCSI_disk.cpp
@@ -591,17 +591,17 @@ static int findNextImageAfter(image_config_t &img,
 
         // keep track of the first item to allow wrapping
         // without having to iterate again
-        if (first_name[0] == '\0' || strcmp(buf, first_name) < 0)
+        if (first_name[0] == '\0' || strcasecmp(buf, first_name) < 0)
         {
             strncpy(first_name, buf, sizeof(first_name));
         }
 
         // discard if no selected name, or if candidate is before (or is) selected
-        if (filename[0] == '\0' || strcmp(buf, filename) <= 0) continue;
+        if (filename[0] == '\0' || strcasecmp(buf, filename) <= 0) continue;
 
         // if we got this far and the candidate is either 1) not set, or 2) is a
         // lower item than what has been encountered thus far, it is the best choice
-        if (candidate_name[0] == '\0' || strcmp(buf, candidate_name) < 0)
+        if (candidate_name[0] == '\0' || strcasecmp(buf, candidate_name) < 0)
         {
             strncpy(candidate_name, buf, sizeof(candidate_name));
         }
--- a/src/ZuluSCSI_disk.cpp
+++ b/src/ZuluSCSI_disk.cpp
@@ -636,10 +636,7 @@ int scsiDiskGetNextImageName(image_config_t &img, char *buf, size_t buflen)
     section[4] = '0' + target_idx;
 
     // sanity check: is provided buffer is long enough to store a filename?
-    if (buflen < MAX_FILE_PATH)
-    {
-        panic("scsiDiskGetNextImageName called with illegal buflen");
-    }
+    assert(buflen >= MAX_FILE_PATH);
 
     if (img.image_directory)
     {
--- a/src/ZuluSCSI_disk.cpp
+++ b/src/ZuluSCSI_disk.cpp
@@ -540,6 +540,16 @@ static void scsiDiskLoadConfig(int target_idx, const char *section)
     memset(tmp, 0, sizeof(tmp));
     ini_gets(section, "Serial", "", tmp, sizeof(tmp), CONFIGFILE);
     if (tmp[0]) memcpy(img.serial, tmp, sizeof(img.serial));
+
+    if (strlen(section) == 5 && strncmp(section, "SCSI", 4) == 0) // allow within target [SCSIx] blocks only
+    {
+        ini_gets(section, "ImgDir", "", tmp, sizeof(tmp), CONFIGFILE);
+        if (tmp[0])
+        {
+            logmsg("-- SCSI", target_idx, " using image directory \'", tmp, "'");
+            img.image_directory = true;
+        }
+    }
 }
 
 int scsiDiskGetNextImageName(image_config_t &img, char *buf, size_t buflen)
--- a/src/ZuluSCSI_disk.cpp
+++ b/src/ZuluSCSI_disk.cpp
@@ -425,8 +425,11 @@ bool scsiDiskOpenHDDImage(int target_idx, const char *filename, int scsi_id, int
 
         return true;
     }
-
-    return false;
+    else
+    {
+        logmsg("---- Failed to load image '", filename, "', ignoring");
+        return false;
+    }
 }
 
 static void checkDiskGeometryDivisible(image_config_t &img)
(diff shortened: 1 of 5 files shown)


src/ZuluSCSI_disk.cpp
```diff
@@ -500,21 +540,160 @@ static void scsiDiskLoadConfig(int target_idx, const char *section)
...
+
+            // keep track of the first item to allow wrapping
+            // without having to iterate again
+            if (first_name[0] == '\0' || strcmp(buf, first_name) < 0)
+            {
+                strncpy(first_name, buf, sizeof(first_name));
+            }
+
+            // discard if no selected name, or if candidate is before (or is) selected
+            if (img.current_image[0] == '\0' || strcmp(buf, img.current_image) <= 0) continue;
```
PetteriAimonen, line 622: Might consider `strcasecmp()` here. That way the order would be more natural "A1, a2, B1, b2" instead of "A1, B1, a2, b2".

src/ZuluSCSI_disk.cpp
```diff
@@ -500,21 +540,160 @@ static void scsiDiskLoadConfig(int target_idx, const char *section)
...
+        {
+            logmsg("-- SCSI", target_idx, " using image directory \'", tmp, "'");
+            img.image_directory = true;
+        }
+    }
 }
 
-// Check if image file name is overridden in config
-bool scsiDiskGetImageNameFromConfig(image_config_t &img, char *buf, size_t buflen)
+int scsiDiskGetNextImageName(image_config_t &img, char *buf, size_t buflen)
```
PetteriAimonen, line 555: I think we may want this to be `scsiDiskGetImageNameFromConfig(image_config_t &img, char *buf, size_t buflen, int image_index)` eventually. That would allow e.g. listing all available images, if we add a I2C display or some GUI tool. It does mean a bit more looping and switching to next image is certainly the common case, so I see the point of starting from current file.

I think the code would be easier to read if there was an internal `findNextImageAfter(const char *dirname, const char *filename, char *buf, size_t buflen)`, and then the `scsiDiskGetImageNameFromConfig()` call that.

In any case I'm ok with the code as it is currently written, this can be worked on later.

src/ZuluSCSI_disk.h
```diff
@@ -32,13 +32,16 @@
...
 #include "ImageBackingStore.h"
+#include "ZuluSCSI_config.h"
 
 extern "C" {
 #include <disk.h>
 #include <config.h>
 #include <scsi.h>
 }
 
+#define IMAGE_INDEX_MAX 9
```
PetteriAimonen, line 43: This could be in `ZuluSCSI_config.h`.

src/ZuluSCSI_disk.cpp
```diff
@@ -500,21 +540,160 @@ static void scsiDiskLoadConfig(int target_idx, const char *section)
...
-    key[3] = '0' + img.image_index;
+    // sanity check: is provided buffer is long enough to store a filename?
+    if (buflen < MAX_FILE_PATH)
+    {
+        panic("scsiDiskGetNextImageName called with illegal buflen");
+    }
+
+    if (img.image_directory)
+    {
+        // is this needed to prevent an inadvertent callback during all these opens?
```
PetteriAimonen, line 570: No, it is not needed. The callback should nulled out after use by any function that sets it. It also only triggers if the data buffer pointer passed to `file.read()` matches the one passed to `set_sd_callback()`.

src/ZuluSCSI_disk.h
```diff
@@ -93,6 +101,9 @@ void scsiDiskCloseSDCardImages();
 bool scsiDiskOpenHDDImage(int target_idx, const char *filename, int scsi_id, int scsi_lun, int blocksize, S2S_CFG_TYPE type = S2S_CFG_FIXED);
 void scsiDiskLoadConfig(int target_idx);
 
+// During loading, checks if a filename is appropriate for further processing as a disk image
```
PetteriAimonen, line 104: Comment could have a bit more detail: checks file extension for ignored file types. Does not check file name prefix.

PetteriAimonen (commented):
Looks good so far, I left a few comments but nothing critical.

I didn't test it yet.

src/ZuluSCSI_disk.cpp
saybur: Perfect, removed.

src/ZuluSCSI_disk.h
saybur: I added some additional language explaining the function. The name was given with the thought that it may be expanded in the future if there are additional things that should cause a file to be ignored, but I agree that it currently is a bit misleading.

src/ZuluSCSI_disk.cpp
saybur: That is definitely more intuitive!

src/ZuluSCSI_disk.cpp
saybur: That's a great point regarding future expandability. The only reservation I have with making `image_index` a parameter is that it exposes more of the image switching logic to callers; I had some trouble getting my head wrapped around what pieces of code affect that variable (which I was making more complicated by the parallel `current_image` name tracker).

This approach will need changes if a later PR adds something like the ability to set images via I2C, SCSI, etc. Perhaps the image switching is something that could be moved to a class to protect implementation details?

You're absolutely right that the function had gotten too long and was hard to read: I broke out the actual directory scanning to a new `findNextImageAfter` call. Hopefully that will help a bit.

saybur:
The feedback thus far is much appreciated, thank you!

saybur:
I've tested this across several platforms and image switching (both from operating systems and external buttons) seems to be working properly. I did not test with huge numbers of images so that may expose some additional problems. Otherwise should be ready for review.

src/ZuluSCSI_disk.cpp
PetteriAimonen: Looks good. I agree that a new class will be good approach in the future.

src/ZuluSCSI_disk.cpp
```diff
@@ -500,21 +540,174 @@ static void scsiDiskLoadConfig(int target_idx, const char *section)
...
 
     char section[6] = "SCSI0";
     section[4] = '0' + target_idx;
 
-    char key[5] = "IMG0";
-    key[3] = '0' + img.image_index;
+    // sanity check: is provided buffer is long enough to store a filename?
+    if (buflen < MAX_FILE_PATH)
+    {
+        panic("scsiDiskGetNextImageName called with illegal buflen");
```
PetteriAimonen, line 641: The `panic()` function appears to be an mbed thing. This is failing to build on GD32 for me.

I think just `assert(buflen >= MAX_FILE_PATH)` would be enough. The address values in assert message need to be resolved with `addr2line` or by using gdb backtrace, but for programmer errors (as opposed to user errors) I consider that ok.

PetteriAimonen (changes requested):
The `panic()` function fails to build for GD32. Otherwise looks good to me.

I'll see about enabling github workflow builds for external pull requests, which would catch problems like this more easily.

src/ZuluSCSI_disk.cpp
saybur: Aah, sorry about that. Fixed. This check could also just be removed, it is partially left over from before separate buffers were being allocated for dirname & filename.

PetteriAimonen (approved):
Ready to merge :)

aperezbios:
Excellent, thanks to both of you. Merged. :)

PR #236, opened 2023-06-06
Fix CD-ROM leadout LBA with BIN/CUE images

PetteriAimonen:
E.g. if the last track has 1000 sectors, the valid LBAs are 0 to 999 and the leadout should start at LBA 1000.

Without this Linux is trying to read 1 sector beyond the end of the image and gives some errors in dmesg about that.

@saybur Can you check if this makes sense?


saybur:
Your description sounds right. I'm not certain why I added the `+1` there, I think it was due to some documentation referring to the lead-out as starting immediately after the last user data track, but it sure seems like the implementation as you wrote it covers that. I can look into it more later today but your change seems logical and is probably safe to merge as-is.

PR #237, opened 2023-06-08
CDROM: Mode page 0x2A, READ TRACK INFORMATION, and GET CONFIGURATION

saybur:
Proposal to include mode page 0x2A support and a minimal implementation of the MMC-1 version of READ TRACK INFORMATION. The goal here is twofold: improve support for drivers expecting to see 0x2A, and work toward suppressing kernel warnings on Linux.

These are lightly tested: the basic commands appear to work under Linux but other operating systems need to be checked. Adding this has not yet eliminated Linux doing a READ(10) on audio data, producing errors. More changes will be needed before this is a candidate for merging.

If this is out of scope or duplicates existing work obviously feel free to close :)

### DIFF
--- a/lib/SCSI2SD/src/firmware/mode.c
+++ b/lib/SCSI2SD/src/firmware/mode.c
@@ -514,6 +514,8 @@ static void doModeSense(
 		idx += sizeof(SequentialDeviceConfigPage);
 	}
 
+	idx += modeSenseCDCapabilitiesPage(pc, idx, pageCode, &pageFound);
+
 	if ((
 			(scsiDev.target->cfg->quirks == S2S_CFG_QUIRKS_APPLE) ||
 			(idx + sizeof(AppleVendorPage) <= allocLength)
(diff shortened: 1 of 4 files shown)


PetteriAimonen:
Good timing, I was about to start adding READ TRACK INFORMATION myself :)

In quick reading looks good to me, but I haven't tested on hardware.
Seems low-risk for merge, just let us know when you consider it ready.

saybur:
Will do. I agree these seem pretty safe; the issue I ran out of time to resolve last night was a mismatch between the new command and GET CAPACITY. The new code was matching a real CD, so I have a suspicion that GET CAPACITY may have an off-by-one error similar to what you caught in #236. I wanted to be sure which one was wrong before marking this as ready. Hopefully will get an answer on that later today.

saybur:
The capacity command appears to be fine after testing.

In very happy news adding GET CONFIGURATION support has stopped the kernel errors.

Testing still needs to be done, but that shouldn't be too complicated at this point. I'll change the state of the PR once that's complete.

PR #243, opened 2023-06-13
Add the ability to delay boot

morio:
Adds the ability to delay booting in milliseconds before and/or after the first time SCSI interface is initialized.

These settings are InitPreDelay and InitPostDelay in the "zuluscsi.ini" file.

This is to fix an issue with the RP2040 boards not working with an MPC3000. Thus a preset has been added for the device name MPC3000 and can be set by using the System="MPC3000" in the "zuluscsi.ini" file. The InitPreDelay is set to 500 milliseconds for the preset.

It should be noted that the InitPreDelay delays the first blink of the board turning on and InitPostDelay hold the LED on.


morio:
@PetteriAimonen could you do a quick code review. I'm not sure if the InitPostDelay would have any practical use as it would initialize the SCSI interface and then just hang for the delay amount.

chickeneps:
Tested directly on a MPC3000 here, works both with "System" and "InitPreDelay", the MPC works with the RP2040 now just fine.

Issue #247, opened 2023-06-17
Issue with CD eject command


guytaylor:
I can't get the CD eject command cycling CD-Rom iso images working using a ZuluSCSI 1.1 (latest firmware) on a Kurzweil K2500

This is my .ini file:

```
[SCSI1]
Product = "CD-ROM Drive"
Type=2 # CDROM drive
IMG0 = img1.iso
IMG1 = img2.iso
IMG2 = img3.iso
EjectButton = 1
```

and I see this in the log:
```
[288ms] -- Opening 'img1.iso' for id:1, specified in zuluscsi.ini
...
[314ms] SCSI ID:1 BlockSize:2048 Type:2 Quirks:0 ImageSize:97336kB
...
[22717ms] Eject button 1 pressed, passing to CD drive SCSI1
[29099ms] Eject button 1 pressed, passing to CD drive SCSI1
[41098ms] Eject button 1 pressed, passing to CD drive SCSI1
[52984ms] Eject button 1 pressed, passing to CD drive SCSI1
[74900ms] Eject button 1 pressed, passing to CD drive SCSI1
```

It is stuck on img1.iso and never changes. 
I have tested all the iso separately and they work.
Should there be a response after the "Eject button 1 pressed" saying the new image name?
I'm trynig to work out if the issue is with the zulu or the k2500

PetteriAimonen:
Can you post a full zululog.txt with debug messages enabled?

PetteriAimonen:
I think the issue may be that we currently switch the image only after the host has polled for status. If the host is not checking the CD-ROM status, that probably won't work. But debug log will tell more.

guytaylor:
I did the following:

1. Powered Zulu
2. Powered K2500
3. Looked at SCSI 1 on the K2500, and I could see disc1.iso
4. Pressed Button 1
5. Browsed away from SCSI 1 and returned to scsi1. Still saw disc1. 
6. Pressed Button 1
7. Browsed away from SCSI 1 and returned to scsi1. Still saw disc1. 
8. Powered down k2500, powered it back up (left Zulu powered on)
9. Looked at SCSI 1 on the K2500, I could see disc1.iso
10. Pressed Button 1
11. Pressed 'SCSI sleep' on K2500, Browsed away from SCSI 1,  and returned to scsi1. Still saw disc1. 
12. Pressed Button 1
13. Browsed away from SCSI 1 and returned to scsi1. Still saw disc1. 

[...]

Issue #230, opened 2023-06-02
ReinsertCDOnInquiry behavior inconsistent with documentation


saybur:
The documented behavior for `ReinsertCDOnInquiry` in the INI file is:

```
#ReinsertCDOnInquiry = 1 # Reinsert any ejected CD-ROM image on Inquiry command
```

This seems to describe a 'CD tray retract' type operation. However, in ZuluSCSI_disk.cpp this setting is used to call cdromReinsertFirstImage(). This prevents multi-image configurations from working on INQUIRY-happy operating systems until the setting is disabled.

For reference, I ran into this on modern Linux: at least on Debian bullseye it likes to issue INQUIRY after every new disk is inserted.

PetteriAimonen:
This functionality was added in #141 based on private discussion with @aperezbios. The main issue to solve was:

> I've had some feedback from a few customers that the current eject behavior of ZuluSCSI firmware is problematic. Can you think of a way to "re-insert" the current/sole CD-ROM ISO upon a bus reset? As it stands right now, the only way to force this to happen is to de-energize (or maybe just hardware re-set) the board.

> CD-ROM is already reinserted after bus reset.
But some hosts won't issue a bus reset on reboot (debug log should tell).

Prior to that, reinsertion happened on BUS RESET only. It appears that some systems didn't issue BUS RESET, though I wasn't able to reproduce this on Mac LC III at the time.

I think closing the tray instead of returning to first image would be consistent with the needs for INQUIRY command, where the problem was that Mac has no software selection for "close tray". But maybe on BUS RESET it still makes sense to restart from first image?

@aperezbios How does this sound:

1) If CD tray is open on `INQUIRY`, close it but don't switch image.
2) On BUS RESET, return to first image.

aperezbios:
That sounds perfectly reasonable to me.

PR #251, opened 2023-06-19
Improve CD image switching logic

PetteriAimonen:
1. Fix ReinsertOnInquiry / ReinsertAfterEject settings getting ignored when specified in `[SCSI]` section

2. Fix misleading debug messages on drives that have only a single image.

3. Do not switch CD image when tray gets closed on Inquiry command (issues #230, #247)

4. Always switch CD image on ejection, no matter if by button or software. Previously switching was on tray close, but only if closed by button or ReinsertAfterEject.

This seems to work reasonably for me, but would benefit from testing on various platforms.

@saybur Can you check if the code logic seems reasonable to you?


saybur:
This seems like a more straightforward approach, looks good!

PR #290, opened 2023-08-31
Set vendor / product id with image file name

phe78:
ZuluSCSI displays the first 16 characters of the image file name (with path and extension removed) instead of fixed DRIVEINFO as defined in ZuluSCSI_config.h


src/ZuluSCSI_disk.cpp
```diff
@@ -229,9 +229,22 @@ static void formatDriveInfoField(char *field, int fieldsize, bool align_right)
     }
 }
 
+// remove path and extension from filename 
+const char* get_image_name(const char* filename)
+{
+    char *j, *r;
+
+    r = (char *)malloc(strlen(filename));
```
PetteriAimonen, line 237: This seems to leak memory. I think it would be better to allocate a buffer for the result and pass a pointer to it as argument, similar to how e.g. `scsiDiskGetNextImageName()` works.

lib/SCSI2SD/include/scsi2sd.h
```diff
@@ -83,7 +83,8 @@ typedef enum
 	S2S_CFG_QUIRKS_APPLE = 1,
 	S2S_CFG_QUIRKS_OMTI = 2,
 	S2S_CFG_QUIRKS_XEBEC = 4,
-	S2S_CFG_QUIRKS_VMS = 8
+	S2S_CFG_QUIRKS_VMS = 8,
+	S2S_CFG_QUIRKS_EMU = 9
```
PetteriAimonen, line 87: I wonder if this feature could be generally useful also - if so, it would be better to use a separate setting like `NameFromImage = 1` instead of tying it to the quirks setting.

src/ZuluSCSI_disk.cpp
```diff
@@ -229,9 +229,22 @@ static void formatDriveInfoField(char *field, int fieldsize, bool align_right)
...
+    strcpy(r, strrchr(filename,'/') + 5);
+    j = strrchr(r, '.');
```
PetteriAimonen, line 239: I think this will crash if the image name does not have an extension (no dot => strrchr returns null pointer).

src/ZuluSCSI_disk.cpp
```diff
@@ -398,7 +425,7 @@ bool scsiDiskOpenHDDImage(int target_idx, const char *filename, int scsi_id, int
         PLATFORM_CONFIG_HOOK(&img);
 #endif
 
-        setDefaultDriveInfo(target_idx);
```
PetteriAimonen, line 401: I think the filename-based name setting could be a separate function instead of putting it into `setDefaultDriveInfo()`. If you call it first, `setDefaultDriveInfo()` will detect that the vendor/product name is already set.

Something like this:

1. Add `bool name_from_image` to `image_config_t`.
2. In `scsiDiskLoadConfig()` use `ini_getbool()` to get the setting from ini file
3. In here, before `setDefaultDriveInfo()` is called, do `if (img.name_from_image) { setNameFromImage(img, filename); }`

PetteriAimonen (changes requested):
I think the memory leak and potential crash in get_image_name() must be fixed before merge.

I would prefer to see this use a separate setting ini-file setting that can be used for any device, instead of tying to EMU quirks, but either is acceptable to me.

PetteriAimonen:
Thanks for contributing, I left a few comments in review.

If you run into any problems or just don't feel like doing changes, just let us know and we'll see how to proceed.

PetteriAimonen (approved):

PetteriAimonen:
Thanks, looks good to merge now.

phe78:
Thanks !

PR #297, opened 2023-09-14
Added ability to change CD audio volume in config

KyleJ61782:
This is a PR that allows the default CD audio volume to be set in the zuluscsi.ini configuration file using the CDAVolume configuration.


PetteriAimonen:
Looks good to merge!

PR #303, opened 2023-09-19
Add Mac sanity checks when Apple quirks mode is enabled

morio:
The main purpose of this code is to emit warnings when Mac quirks are turned on and the image(s) are disks.

The warnings are emitted if any of the following checks fail:
    - The disk image contain Apple Magic signature
    - The HFS block size is correct
    - The Mac SCSI driver is smaller than or equal to the max size
    - The MAC SCSI driver is smaller than or equal to the image size
    - The Lido driver is not installed

The original implementation put the sanity checks under platform code. Since none of the code is platform specific, it has been moved to the root src directory so all build targets can take advantage of the sanity checks.

Initial code authored by Eric Helgeson and taken from: https://github.com/BlueSCSI/BlueSCSI-v2/commit/cc1ff5f4d03ce976d01ead09cee1106ce0acae08 https://github.com/BlueSCSI/BlueSCSI-v2/commit/d00ebe016b8c7ef4777c31cef9cd002043de7d46

### DIFF
--- a/src/ImageBackingStore.cpp
+++ b/src/ImageBackingStore.cpp
@@ -1,6 +1,6 @@
 /**
- * Portions - Copyright (C) 2023 Eric Helgeson
  * ZuluSCSI™ - Copyright (c) 2022-2023 Rabbit Hole Computing™
+ * Portions - Copyright (C) 2023 Eric Helgeson
  *
  * This file is licensed under the GPL version 3 or any later version. 
  *
--- a/src/ImageBackingStore.cpp
+++ b/src/ImageBackingStore.cpp
@@ -1,8 +1,8 @@
 /**
- * ZuluSCSI™ - Copyright (c) 2022 Rabbit Hole Computing™
+ * Portions - Copyright (C) 2023 Eric Helgeson
+ * ZuluSCSI™ - Copyright (c) 2022-2023 Rabbit Hole Computing™
  *
  * This file is licensed under the GPL version 3 or any later version. 
- * It is derived from disk.c in SCSI2SD V6
  *
  * https://www.gnu.org/licenses/gpl-3.0.html
  * ----
@@ -143,6 +143,12 @@ bool ImageBackingStore::isWritable()
 bool ImageBackingStore::isRom()
 {
     return m_isrom;
+}
+
+
+bool ImageBackingStore::isRaw()
+{
+    return m_israw;
 }
 
 bool ImageBackingStore::close()
--- a/src/QuirksCheck.cpp
+++ b/src/QuirksCheck.cpp
@@ -25,6 +25,7 @@
 #include "ZuluSCSI_disk.h"
 #include "ZuluSCSI_log.h"
 #include "QuirksCheck.h"
+#include <assert.h>
 #include <stdint.h>
 
 
--- a/src/QuirksCheck.cpp
+++ b/src/QuirksCheck.cpp
@@ -34,33 +34,29 @@ static bool isValidMacintoshImage(image_config_t *img)
     const char apple_magic[2] = {0x45, 0x52};
     const char block_size[2]  = {0x02, 0x00};  // 512 BE == 2
     const char lido_sig[4] = {'C', 'M', 'S', '_' };
-    uint8_t tmp[4] = {0};
-
+    uint8_t tmp[SD_SECTOR_SIZE];
     // Check for Apple Magic
-    img->file.seek(0);
-    img->file.read(tmp, 2);
-    if(memcmp(apple_magic, tmp, 2) != 0)
+    img->file.seek(0); 
+    img->file.read(tmp, SD_SECTOR_SIZE);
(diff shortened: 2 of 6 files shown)


src/QuirksCheck.cpp
```diff
@@ -0,0 +1,120 @@
...
+        if((img->scsiId & S2S_CFG_TARGET_ID_BITS) == 7)
+        {
+          logmsg("---- WARNING: Quirks set to Apple so can not use SCSI ID 7!");
+        }
+    }
+}
+
+void quirksCheck(image_config_t *img)
+{
+    macQuirksSanityCheck(img);
```
PetteriAimonen, line 119: I think it would make sense to have `if (img->quirks == S2S_CFG_QUIRKS_APPLE) { macQuirksSanityCheck(img); }` here, instead of checking it on line 96.

That way there won't be any spurious debug messages when the quirks aren't enabled for the device.

src/QuirksCheck.cpp
```diff
@@ -0,0 +1,120 @@
...
+
+// Called from ZuluSCSI_disk after image is initalized.
+static void macQuirksSanityCheck(image_config_t *img)
+{
+    if(ini_getbool("SCSI", "DisableMacSanityCheck", false, CONFIGFILE))
+    {
+        dbgmsg("---- Skipping Mac sanity check due to DisableMacSanityCheck");
+        return;
+    }
+    if(img->file.isRom() || img->file.isRaw())
```
PetteriAimonen, line 91: ImageBackingStore constructor converts contiguous files to raw mappings. So this check will skip most image files.

Is there a particular reason to skip the check for raw mappings?

src/QuirksCheck.cpp
```diff
@@ -0,0 +1,120 @@
...
+{
+    bool result = true;
+    const char apple_magic[2] = {0x45, 0x52};
+    const char block_size[2]  = {0x02, 0x00};  // 512 BE == 2
+    const char lido_sig[4] = {'C', 'M', 'S', '_' };
+    uint8_t tmp[4] = {0};
+
+    // Check for Apple Magic
+    img->file.seek(0);
+    img->file.read(tmp, 2);
```
PetteriAimonen, line 41: I think the reason why raw mappings are excluded in the `macQuirksSanityCheck()` is these unaligned accesses, which won't work because raw access must be in whole SD blocks. This ends up excluding any continuous image file.

This could be modified to be `uint8_t tmp[SD_SECTOR_SIZE]; img->file.seek(0); img.file.read(tmp, SD_SECTOR_SIZE);` and then comparing against specific bytes in the tmp array.

src/ImageBackingStore.cpp
```diff
@@ -1,8 +1,8 @@
 /**
- * ZuluSCSI™ - Copyright (c) 2022 Rabbit Hole Computing™
+ * Portions - Copyright (C) 2023 Eric Helgeson
+ * ZuluSCSI™ - Copyright (c) 2022-2023 Rabbit Hole Computing™
```
PetteriAimonen, line 3: Order of copyright notices seems wrong, most of the code in this file is from Rabbit Hole Computing.

src/ImageBackingStore.cpp
```diff
@@ -1,8 +1,8 @@
...
  *
  * This file is licensed under the GPL version 3 or any later version. 
- * It is derived from disk.c in SCSI2SD V6
```
PetteriAimonen, line 5: I agree, ImageBackingStore.cpp doesn't have SCSI2SD code in it, the notice has been accidentally copied over when it was separated from ZuluSCSI_disk.cpp.

PetteriAimonen (changes requested):
I think currently this doesn't check any continuous image file, because they get converted to raw mapping.

morio:
Thanks for the feedback @PetteriAimonen, I'll implement the changes.

src/QuirksCheck.cpp
morio: I think BlueSCSI wanted to skip raw pass through and neither of us understood what m_israw actually means. Getting rid of it.

morio: I've got it seeking and reading on SD_SECTOR_SIZE blocks, that means we can actually do a sanity check for ROM images too?

morio:
@PetteriAimonen I think I addressed all the issues. I also removed the isRom() check because all reads and seeks are on SD_SECTOR_SIZE boundaries.

aperezbios:
@morio I noticed that the CI builds for ZuluSCSI V1.x are failing:
```
ZuluSCSIv1_0           FAILED    00:00:19.653
ZuluSCSIv1_0_mini      FAILED    00:00:01.305
ZuluSCSIv1_1           FAILED    00:00:01.583
```

PetteriAimonen (changes requested):
Code looks good now.

The build failure is a missing `#include <assert.h>` in QuirksCheck.cpp.

PR #313, opened 2023-10-12
ZuluSCSI v1.1 ODE and ZuluSCSI v1.2 support

morio:
This introduces support for two new boards in the same ZuluSCSI v1.1 firmware file
 - ZuluSCSI v1.1 ODE (Optical Drive Emulator)
 - ZuluSCSI v1.2 with direct mode

It also allows whole ZuluSCSI v1.x line to send logging over a serial USB com port

The ZuluSCSI v1.1 ODE has an audio out header that transmits CD audio over a builtin DAC just like a normal CDROM drive.
This is useful for not only audio CD images but also mixed mode CDs like some older games that played CD audio as the sound track and had the installer on the data track.

The ZuluSCSI v1.2 has a direct mode that currently treats the SD card as a single block device. Its onboard rotary dip switch selects the device type: HDD(fixed), Tape (sequential), CD-ROM (optical), or RE/Floppy (removable). Another normal dip switch selects which SCSI ID it should present itself on the SCSI bus.

With the added support the firmware  ZuluSCSIv1_1_\*.bin file has been renamed ZuluSCSIv1_1_plus_\*.bin.


PetteriAimonen:
I'm not familiar enough with the hardware differences for in-detail review, but I had a quick look and it all seems very reasonable.

PR #375, opened 2024-02-05
Siemens EWSD patch for ZuluSCSI

Eversberg:
Supports MDD and MOD as if it is an original drive. Works at first try, no additinal reset required to boot from drive.

EWSD is a class 5 telephone switching system, made in Germany.


morio:
@Eversberg Looks good, moving the pull request to a feature branch to change `unitAttention_stop` to `unitAttentionStop` then pull it to main.
Thank you for contributing!

aperezbios:
@Eversberg thank you for your contribution, it is appreciated!

PR #379, opened 2024-02-10
Added Xebec Support

peclark1:
Added Xebec support by porting code from SCSI2SD V5 firmware. This code supports the First Class Peripherals Sider hard disk for Apple II


src/ZuluSCSI_log_trace.cpp
```diff
@@ -100,6 +100,8 @@ static const char *getCommandName(uint8_t cmd)
         case 0xA8: return "Read12";
         case 0xC0: return "OMTI-5204 DefineFlexibleDiskFormat";
         case 0xC2: return "OMTI-5204 AssignDiskParameters";
+        case 0xE0: return "RAM Diagnostic";
+        case 0xE4: return "Drive Diagnostic";
```
morio, line 104: @peclark1 
I think prefixing these string with `Xebec ` would make them more inline with the two Xebec commands listed above.
https://github.com/ZuluSCSI/ZuluSCSI-firmware/blob/8671a0b6c40437300d59e0e5f4ed34b5547eb14a/src/ZuluSCSI_log_trace.cpp#L52-L53

`"Xebec RAM Diagnostic"`
`"Xebec Drive Diagnostic"`

src/ZuluSCSI_disk.cpp
morio, line 1: @peclark1 
Does it work with a Xebec controller if you move these to vendor.c after here:
https://github.com/ZuluSCSI/ZuluSCSI-firmware/blob/8671a0b6c40437300d59e0e5f4ed34b5547eb14a/lib/SCSI2SD/src/firmware/vendor.c#L75-L82

Like this:
```C++
 else if (command == 0x0F && 
 	scsiDev.target->cfg->quirks == S2S_CFG_QUIRKS_XEBEC) 
 { 
 	// XEBEC S1410, WD100x: "Write Sector Buffer" 
 	scsiDev.dataLen = scsiDev.target->liveCfg.bytesPerSector; 
 	scsiDev.phase = DATA_OUT; 
 	scsiDev.postDataOutHook = doWriteBuffer; 
 } 
  else if (command == 0xE0 && 
 	  scsiDev.target->cfg->quirks == S2S_CFG_QUIRKS_XEBEC)
  {
      // RAM Diagnostic
      // XEBEC S1410 controller
      // http://bitsavers.informatik.uni-stuttgart.de/pdf/xebec/104524C_S1410Man_Aug83.pdf
      // Stub, return success
  }
  else if (command == 0xE4 && 
 	  scsiDev.target->cfg->quirks == S2S_CFG_QUIRKS_XEBEC)
  {
      // Drive Diagnostic
      // XEBEC S1410 controller
      // Stub, return success
  }   
```
That way they would be in the same place with the other Xebec commands.
If you have a system to verify the code changes with, that would be great.

morio (commented):
@peclark1 
Here are some suggestions for changes to the pull request. 
I don't have a system to test them with, but if you do, could you verify they work?

src/ZuluSCSI_disk.cpp
peclark1: sure, I'll build and give it a test.

peclark1: I totally missed the SCSI2SD source under the lib directory - feel like an idiot.   I'll move my changes over there.

peclark1:
Ok, changed made as requested.   Tested both Initializing/Formatting new hard disk, as well as accessing partitions (prodos and dos).

morio:
@peclark1 thanks for the great work!

PR #380, opened 2024-02-13
Initiator: allow bus to settle after REQ

morio:
Fixes instability in initiator transfers.

@androda changes in pull request: https://github.com/BlueSCSI/BlueSCSI-v2/pull/92 applied to ZuluSCSI


morio:
This shouldn't be needed if the device follows the SCSI spec and it does reduce the speed of copying.
Maybe revisit with an fix that can be controlled via a setting in the zuluscsi.ini file.

PR #391, opened 2024-03-08
Daynaport Amiga patch from Rob Smith

reinauer:
- **Free up some space for Daynaport**
- **Support Amiga driver for DaynaPort**


aperezbios:
Excellent. Thank you for your contribution!

PR #406, opened 2024-03-21
Implement the SCSI side of a SCSI control and data transfer mechanism

morio:
Code was taken from BlueSCSI's https://github.com/BlueSCSI/BlueSCSI-v2/pull/79
and modified for use in ZuluSCSI.

Changes include the vendor page so the BluSCSI toolbox transfer
application will not work with this build.

Also changed which files are exclude from the toolbox's CD changer.
Now subdirectories, common compressed file and document extensions are excluded from the list.

Moved scsiToolboxCBDLen code from Toolbox.cpp into vendor.c so CDB vendor command lengths have visibility in the same file.


src/Toolbox.cpp
```diff
@@ -0,0 +1,412 @@
...
+    while (file.openNext(&dir, O_RDONLY))
+    {   
+        uint8_t isDir = file.isDirectory() ? 0x00 : 0x01;
+        int len = file.getName(name, MAX_FILE_PATH);
+        if (len > MAX_MAC_PATH)
+            name[MAX_MAC_PATH] = 0x0;
+        uint64_t size = file.fileSize();
+        file.close();
+        if (!toolboxFilenameValid(name, isCD))
+            continue;
```
nielsmh, line 117: This `toolboxFilenameValid` check in `onListFiles` is different from the one in `doCountFiles`. When listing files, the filename is truncated to MAX_MAC_PATH characters before checking for being value, but when counting files the filename is not truncated. So the count could include files that would be excluded in the listing, if their filenames are too long and the extension gets cut off.

src/Toolbox.cpp
```diff
@@ -0,0 +1,412 @@
...
+
+        for(int i = 2; i < (MAX_MAC_PATH + 1 + 2); i++) {   // bytes 2 - 34
+            file_entry[i] = name[c++];
+        }
+        file_entry[35] = 0; //(size >> 32) & 0xff;
+        file_entry[36] = (size >> 24) & 0xff;
+        file_entry[37] = (size >> 16) & 0xff;
+        file_entry[38] = (size >> 8) & 0xff;
+        file_entry[39] = (size) & 0xff;
+        memcpy(&(scsiDev.data[ENTRY_SIZE * index]), file_entry, ENTRY_SIZE);
```
nielsmh, line 132: Buffer overrun if the directory contains more than 102 valid images.

src/Toolbox.cpp
nielsmh: `get_file_from_index` also does not truncate the filenames, so it counts indexes different from `onListFiles` too.

PR #420, opened 2024-05-09
Various Toolbox fixes

nielsmh:
This PR addresses some issues I found with the Toolbox implementation from #406, while implementing my own client for DOS/ASPI.

Note that these changes are only tested with my own ZuluSCSI RP2040 based hardware, on a single machine with an Adaptec host controller using ASPI8DOS.sys driver. The client software I have developed/tested with is published at <https://github.com/nielsmh/escsitoolbox>.
I do not own a Mac capable of running the original Toolbox software, so I don't know if these changes affect compatibility with that.


aperezbios:
@nielsmh thank you for this valuable contribution. This is awesome :) We will merge it after testing on our end has concluded.

morio:
Other 0xD8 vendor commands pass testing
Thanks for the PR

Issue #432, opened 2024-06-29
Apple quirks overrides toolbox calls


saybur:
If both the toolbox and S2S_CFG_QUIRKS_APPLE are active the CD-ROM command handler gets 0xD8/D9 before the toolbox handler is called. Obviously some command conflict is unavoidable here but it looks like the intended behavior was to have the toolbox go first:

<https://github.com/ZuluSCSI/ZuluSCSI-firmware/blob/8b541a8d61759c936f4516822db2489b402db8f4/lib/SCSI2SD/src/firmware/vendor.c#L202-L210>

PR #434, opened 2024-07-03
Adjust logic for supported vendor commands

morio:
There isn't any functional change in this fix, just a rearrangement with if else statements to address this issue: https://github.com/ZuluSCSI/ZuluSCSI-firmware/issues/432 Also a for loop wasn't initialized in `src/ZuluSCSI.cpp`, set the counter to 0.

### DIFF
--- a/lib/SCSI2SD/src/firmware/scsi.c
+++ b/lib/SCSI2SD/src/firmware/scsi.c
@@ -599,8 +599,15 @@ static void process_Command()
 	{
 		enter_Status(CONFLICT);
 	}
+	// Handle Toolbox commands, overriding other vendor commands if enabled
+	else if (scsiToolboxEnabled() && scsiToolboxCommand())
+	{
+		// already handled
+	}
 	// Handle odd device types first that may override basic read and
 	// write commands. Will fall-through to generic disk handling.
+	// Some device specific vendor commands are located here instead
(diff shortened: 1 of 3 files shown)


saybur:
Sorry, I wasn't being very helpful with my link in the issue. If that linked comment's intent is right, then I think the problem is actually here:

https://github.com/ZuluSCSI/ZuluSCSI-firmware/blob/3b81282639c2e9df3eb45a1b6581ba7624287ce3/lib/SCSI2SD/src/firmware/scsi.c#L604

Where the CD-ROM handler is being called and resolves the command before the vendor handler is called:

https://github.com/ZuluSCSI/ZuluSCSI-firmware/blob/3b81282639c2e9df3eb45a1b6581ba7624287ce3/lib/SCSI2SD/src/firmware/scsi.c#L639

morio:
Ahh yeah, I see the issue. Thanks!
I could add a !scsiToolboxEnabled() to the three CD commands. 
It probably make more sense to move 
https://github.com/ZuluSCSI/ZuluSCSI-firmware/blob/3b81282639c2e9df3eb45a1b6581ba7624287ce3/lib/SCSI2SD/src/firmware/vendor.c#L174-L177
into the scsi.c above the specific device handling.

PR #468, opened 2024-11-02
Ultra SCSI (Fast20) support with reclocking the RP2040 and RP2350 line of boards and user editable reclocking timings

morio:
This contains changes allowing the user to set `SpeedGrade` to `TurboMax` in the `zuluscsi.ini` file and get speeds up to 17MB/s reads with RP2040 based boards and over 10MB/s writes with a quality Class 10 SD card.

Timings can be fined tuned with `zuluscsi_timings.ini` as the Turbo speeds grades have not been fully flushed out.
Users can open the USB serial console and hit the 'm' key followed by the 'y' key to boot into mass storage mode, change either `zuluscsi.ini` or `zuluscsi_timings.ini`, eject the mass storage device from the host computer and have their timing changes loaded into the ZuluSCSI without needed to eject the SD card or power off the ZuluSCSI.

In `zuluscsi_timings.ini` file the sync period and sync offset can be set to the value negotiated by the SCSI host so when restarting via the USB serial console you don't loose the sync on the device side. Helpful for speed testing with command like `dd` in Linux.

Overclocking the RP2040 to our limit of 250MHz is beyond the factory specs and the long term side effect are unknown.
You can google https://www.google.com/search?q=overclocking+the+rp2040+to+250MHz for more information


morio:
Binaries from the CI for the initial pull request
[ZuluSCSI Turbo binaries.zip](https://github.com/user-attachments/files/17605185/ZuluSCSI.Turbo.binaries.zip)

Run vcocalc.py here: https://www.online-python.com/LV1oJhiXe3
Enter in your target system clock as floating point number as MHz in the "Command Line Arguments" textbox to get your PLL settings

PR #478, opened 2024-12-03
Switch activity LED blinking from delay to polling

morio:
The purpose behind this change is to stop having USB console logging be jittery because of blinking function when there is no SD card inserted. This was due to the blinking function being delay based. Switched the blink function to polling and checks when blinking should occur with deadlines.


aperezbios:
Thanks for the feedback, @PetteriAimonen. Merging with updates.

PR #501, opened 2025-01-22
Present images as USB storage on RP2 platform

ZigZagJoe:
Adds the ini configuration option USBMassStoragePresentImages.

If set, MSC will present all configured eligible images as LUNs. If no images are available, it will fall back to presenting the SD card as storage. Eligibility is defined by if the image can be expected to contain a partition table or not; we can't put an optical device as a removable drive.

**NOTE**: This will fail to compile as currently presented. This is due to the platform MSC needing to pull in ZuluSCSI_disk.h in order to understand the image_config_t structure.  ZuluSCSI_disk.h tries to pull in CUEParser.h but fails to compile as PlatformIO does not seem to pull in libraries while compiling the platform specific code.
Not sure of the best way to resolve this. 

UI could be improved: A method to switch between presenting SD card and images at runtime would be useful. Otherwise, it is not possible to switch back to card reader mode if you set the .ini setting using USB in the first place.

GDS platform uCs are not currently supported. It would be possible to do so, however the libraries provided by gigadevice would need to be restructured in order to track capacity/block size/eject status on a per-LUN basis rather than global.

### DIFF
--- a/lib/ZuluSCSI_platform_GD32F205/ZuluSCSI_platform_msc.cpp
+++ b/lib/ZuluSCSI_platform_GD32F205/ZuluSCSI_platform_msc.cpp
@@ -117,6 +117,11 @@ bool platform_sense_msc() {
   return false;
 }
 
+void platform_set_msc_image_mode(bool image_mode) {
+  if (image_mode) logmsg("Warning: USB Image mounting not supported on this platform!");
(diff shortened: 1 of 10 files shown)


morio (approved):
Looks good, surprised how little code was needed.
Thanks Joe!

PR #512, opened 2025-02-11
RP2040: Fix 10s delay before USB serial port became available on Windows

PetteriAimonen:
Since commit 196dd9f69db, RP2040 ZuluSCSI_platform_msc.cpp function tud_msc_get_maxlun_cb() was returning 0 if MSC mode was not active. The USB descriptors still show MSC functionality so it is valid for the host to query the lun count.

This caused TU_VERIFY(maxlun); fail in tinyusb/src/class/msc/msc_device.c:378 which then stalled the endpoint.

The endpoint stall caused a delay before the CDC serial port part of the USB compound device was initialized.

Fixed by always returning at least 1 lun, and then reporting that lun as media not present.

image

### DIFF
--- a/lib/ZuluSCSI_platform_RP2MCU/ZuluSCSI_platform_msc.cpp
+++ b/lib/ZuluSCSI_platform_RP2MCU/ZuluSCSI_platform_msc.cpp
@@ -245,7 +245,16 @@ extern "C" uint8_t tud_msc_get_maxlun_cb(void)
   MSCScopedLock lock;
   if (g_msc_initiator) return init_msc_get_maxlun_cb();
 
-  return g_MSC.lun_count; // number of LUNs supported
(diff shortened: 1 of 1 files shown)


ZigZagJoe:
Unfortunately this has introduced a bug: Windows will only ask the number of LUNs at enumeration, so when MSC has fully initialized and knows how many images it will be presenting that count has changed but windows never queries again.

I worked around a sort-of similar issue in another project by issuing a unit not ready sense code once.

On one call to tud_msc_test_unit_ready_cb(uint8_t lun)
call tud_msc_set_sense(lun, SCSI_SENSE_NOT_READY, 0x3a, 0x00) and return false. 
Subsequent calls return true as normal. 

AFAIK windows incessantly polls the the device, which is why this works. But it's been a while.

PR #541, opened 2025-03-24
Playback based on track/index

morio:
Implemented Play Audio Track Index SCSI command. This command was encountered on a Konami arcade.


aperezbios (approved):
This is a ZuluSCSI Blaster-specific enhancement

PR #559, opened 2025-04-24
DaynaPORT SCSI/Link support for ZuluSCSI Blaster

morio:
This commit enables support for Raspberry Pi's new RM2 Radio Module. The RM2 module contains the same Infineon transceiver ([CYW43439](https://www.infineon.com/cms/en/product/wireless-connectivity/airoc-wi-fi-plus-bluetooth-combos/wi-fi-4-802.11n/cyw43439/)) as used in the Pico W and Pico 2W, and supports IEEE 802.11 b/g/n wireless LAN.

RM2 works with an audio compatible clock speed of around 155MHz. ZuluSCSI Blaster now initializes at ~155MHz. If no RM2 is detected, it then re-clocks to 200.5MHz.

* Switched the variant in the board `zuluscsi_blaster.json` file to be rpipico2w so the proper static libraries with wifi support get linked.

* Turn on the RM2 module's LED and the Pico 1W/2W LEDs (not the activity LED) early on at firmware initialization.

This indicates a couple of things to the user
 - The devices are setup in network mode
 - The communication interface between the ZuluSCSI and the wireless module is functional

The RM2 and Pico W LED turns off when either the network connection is attempted, in which case the LED goes through some blinking sequences, or there is no network device configured on a SCSI ID.

### DIFF
--- a/boards/zuluscsi_blaster.json
+++ b/boards/zuluscsi_blaster.json
@@ -9,7 +9,7 @@
         },
         "core": "earlephilhower",
         "cpu": "cortex-m33",
-        "extra_flags": "-DARDUINO_GENERIC_RP2350 -DARDUINO_ARCH_RP2040 -DUSBD_MAX_POWER_MA=500 ",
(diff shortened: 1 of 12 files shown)


aperezbios:
RM2 standalone
RM2 Carrier bottom
RM2 on ZuluSCSI Blaster carrier

PR #614, opened 2025-08-05
Added kiosk mode functionality for museum and demonstration setups

fstark:
Added a kiosk mode to ZuluSCSI. Added a note to the README.

Any file ending with `.ori` will be copied into a new file at startup. `HD20_512.hda.ori` => `HD20_512.hda`

As suggested in https://github.com/ZuluSCSI/ZuluSCSI-firmware/discussions/611 copy is done using the 64KB scsi buffer. However, there are cases where the multi-block I/O fails, so I had to implement a retry logic on read. If this feels to complex, let me know. The 64Kb buffer gives around 2.2Mb/s on my hardware, while a single 8Kb buffer is 1.8Mb/s, so it is not insignificant.

Later, I may look into making a copy-on-write version, as the current implementation is not suitable for large disks.

### DIFF
--- a/src/ZuluSCSI.cpp
+++ b/src/ZuluSCSI.cpp
@@ -944,6 +944,9 @@ static bool poll_sd_card()
 #endif
 }
 
+// Forward declaration for kiosk restore function
+static void kiosk_restore_images();
+
 static void init_eject_button()
 {
   if (platform_has_phy_eject_button() &&  !g_scsi_settings.isEjectButtonSet())
@@ -1065,6 +1068,7 @@ static void zuluscsi_setup_sd_card(bool wait_for_card = true)
       delay(boot_delay_ms);
     }
     platform_post_sd_card_init();
+    kiosk_restore_images();
     reinitSCSI();
 
     boot_delay_ms = cfg->initPostDelay;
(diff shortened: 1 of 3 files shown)


src/ZuluSCSI.cpp
```diff
@@ -1222,3 +1226,210 @@ extern "C" void zuluscsi_main_loop(void)
...
+  if (err_count > 0)
+    logmsg("Kiosk restore: Read completed with ", err_count, " errors");
+
+  return total_read;
+}
+
+// Kiosk mode: Restore image files from .ori backups for museum installations
+static void kiosk_restore_images()
+{
+  logmsg("Kiosk 0.3 restore: Checking for .ori backup files to restore");
```
PetteriAimonen, line 1276: I think this log message is unnecessary (currently it always gets printed, whether there are any .ori images or not).

PetteriAimonen: It's easiest if you can push a commit to your main that removes the log message, it will then automatically be part of this pull request. Otherwise I think I can only do it after the PR is merged (can't edit it directly).

fstark:
> Looks good to me. I agree that this will be easy to maintain.
> 
> I agree that the SDFat (or possibly our SDIO library) error handling is sometimes a bit weird when transfer fails in the middle.

The bug is that the file position is incremented for each successful block, but -1 is returned for the whole operation. I know the exact location of the problem and I could send a fix upstream. However, this may have adverse consequences if anyone is relying on this behavior (I may try to do this however).

> That should be very rare though if the SD card communication is working ok.

Unfortunately it is systematic on my setup, 1 get between 1 and 3 errors on a 40Mb transfer. The 512 bytes retry handles it perfectly.

> For reference, I get 4.3 MB/s restore on ZuluSCSI RP2040 with a fast SD card, and 9.5 MB/s on ZuluSCSI Blaster.

Your hardware is way better than mine :-). You may want to update the README with the number range.

> Regarding future development, if you are interested in implementing copy-on-write. I think easiest way would be to add to `ImageBackingStore` a dynamically allocated bitmap buffer. This would then identify whether a block has been written since boot, and read it either from .ori or the main image file. But for large images the amount of RAM used for the bitmap would grow quite large. It might be best to scale the block size based on image size to keep constant RAM usage even for large files.

Yes, the COW implementation I plan to do is very similar to your description. I envision a 1Kb bitmap per file (could be a config option), so there would be up to 8192 chunks for the image (chunk size being a multiple of 512). That would be chunk size of 5120 bytes for a 40Mb disk, and 256K for 2Gb. There would be a penalty hit at the first write to a new chunk, but in a museum setup, I don’t see random poor write performance being an issue.

src/ZuluSCSI.cpp
fstark: Done (sorry for being a n00b a pull requesting). Made the change + updated the README to indicate a range of speed instead of my sluggish 2Mb/s.

PR #635, opened 2025-08-18
CLEAR_ROM file was crashing the ZuluSCSI

morio:
Placing a CLEAR_ROM file on the root directory on the SD card to clear the ROM drive was crashing the ZuluSCSI. Best guess is it was trying to read beyond an SRAM boundary as there was a read overrun because the write buffer was smaller than the byte count to be written.

Creating a zeroed buffer the same size as the sector size of the flash corrects the issue.


PetteriAimonen:
Yeah, looks like this is the `CLEAR_ROM` equivalent of e9944073. Fix looks good.

PR #642, opened 2025-08-22
Fix scsiLogDataOut logging commands

morio:
Commit: https://github.com/ZuluSCSI/ZuluSCSI-firmware/commit/7ab702172375f86b02ad37435abc194fc737bc8d broke logging SCSI commands and would log SCSI OUT reads that were null.

This fix is to not call scsiLogDataOut in scsiFinishRead when called with NULL buffer. And in scsiRead after a scsiFinishRead is called with a NULL buffer call scsiLogDataOut so the COMMAND and/or OUT log message will be captured.


PetteriAimonen (approved):
Looks like a correct fix.

The scsiStartRead/scsiFinishRead and same for write API turned out to have some hidden assumptions that I've never really thought about. But considering it is only called from two places, I hope there aren't much more lurking anymore :)

PR #651, opened 2025-09-02
First release of Zulu Control support for ZuluSCSI Blaster

guytaylor:
Features
- Browsing all types
	- ImgDir folders - This also has caching and categories
	- Standard folders e.g. hd0 (same as ImgDir)
	- IMGx files
	- prefix files
- Copy progress screen for:
	- Initiator
	- ROM Copy
	- Kiosk Mode
	- Create Image
- SCSI Map screen
- Info for device
- 2 styles of browse screen
- Browse Type Selector
- Categories

Bugs
- Rotary skips clicks sometimes and doesn't keep up is turned quickly (flat view I think)

Things to note
- redrawing the display is a slow
	- if you are on a screen with a scroller it may slow down transfer etc by approx 10-20%
	- to avoid
		- go to a screen (e.g. home) which has no scrollers
		- keep file names / paths short so they don't scroll
	- (can I multithread for the UI? Need to know how on the RP)
- I've extended MAX_FILE_LEN to 260
 - as there can be folders on the SD now, 64 really was limiting 
 
Things which need testing still (I haven't tested these at all)
- Bin/Cue images
- if you have a non-browsable image in a list with browsables
- when there is a RAW image/SD

Things not supported yet
- MSC mode


aperezbios:
@guytaylor the CI failed to build this due to what appears to be a missing dependency/header. See https://github.com/ZuluSCSI/ZuluSCSI-firmware/actions/runs/17401509516/job/49448556343#step:5:2821 for the full details. Here's an excerpt:
```
lib/ZuluSCSI_platform_RP2MCU/ui.cpp:3:10: fatal error: ..\src\ui.h: No such file or directory

*******************************************************************
* Looking for ..\src\ui.h dependency? Check our library registry!
*
* CLI  > platformio lib search "header:..\src\ui.h"
* Web  > https://registry.platformio.org/search?q=header:..%5Csrc%5Cui.h
*
*******************************************************************

    3 | #include "..\src\ui.h"
      |          ^~~~~~~~~~~~~
compilation terminated.
```

guytaylor:
I think it doesn't like my ".." in paths. So I'm changing my includes to not have any. Hopefully that fixes it!

guytaylor:
Argh, same issue, different file. filename casing...

PR #657, opened 2025-09-06
Initial support for hardware UI control board on RP2040 and Wide

morio:
Using dynamic memory adjustment to the scsiDev.data buffer allows the RP2040 to fit either the controller board objects or the S/PDIF audio buffers. The length of the scsiDev.data buffer for SCSI buffering is decreased which does cause throughput degradation, but if neither the controller board nor the S/PDIF interfaces are used then the RP2040 will operate with a full sized scsiDev.data buffer at its original speed.

### DIFF
--- a/lib/ZuluSCSI_UI_RP2MCU/BrowseMethod.h
+++ b/lib/ZuluSCSI_UI_RP2MCU/BrowseMethod.h
@@ -1,3 +1,24 @@
+/**
+ * Copyright (c) 2025 Guy Taylor
+ *
+ * ZuluSCSI™ firmware is licensed under the GPL version 3 or any later version.
+ *
+ * https://www.gnu.org/licenses/gpl-3.0.html
+ * ----
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+**/
+
 #if defined(CONTROL_BOARD)
 
 #ifndef BROWSEMETHOD_H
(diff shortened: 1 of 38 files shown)


morio:
@guytaylor I've got this working for the RP2040, and it does still work with the Blaster. But I'm running into an issue where the SCSI Map is showing garbage IDs on boot, if I reinsert the SD card it loads correctly. This is confirmed to be the case on the Blaster too.

Any ideas?
scsi-map

 The code is in this branch https://github.com/ZuluSCSI/ZuluSCSI-firmware/tree/feature/controller-on-rp2040

I moved all your code to a separate library, for better memory management with the linker 
https://github.com/ZuluSCSI/ZuluSCSI-firmware/tree/feature/controller-on-rp2040/lib/ZuluSCSI_UI_RP2MCU

morio:
@guytaylor thanks, I'll look into it on Monday.
If I remember correctly my zuluscsi.ini file had just "System = Mac"
I wasn't using any special  image directories (maybe that is the cause?)
I had two HDDs on SCSI id 0 and 1 in the root directly along with and NE4.img for a wifi module.
And a bin/cue on SCSI ID 3.

Something like
```
root of SD card
|- hd0-mac-system.img
|- hd1-second-drive.img
|- ne4.img
|- cd3-descent.bin
|- cd3-descent.cue
```

guytaylor:
Ok, this should fix it:

in control.cpp

```
extern "C" void sdCardStateChanged(bool absent)
{
    g_sdAvailable = !absent;

    initDevices();

    if (!g_controlBoardEnabled)
    {
        return;
    }

    if (absent) // blank the device map
    {
        stateChange();
    }
}
```

Move  initDevices(); up out of the if to the op of the method.
I'll see if this has any other impact, but so far so good with your sd setup :)

guytaylor:
I just looked over the change. Looks good to me!

morio:
@PetteriAimonen I reversed sharing the scsiDev.data buffer with RP2040 code. I did keep the concept though by using the extra space you freed up with commit https://github.com/ZuluSCSI/ZuluSCSI-firmware/commit/4e1e3a7fc4bc6ad183ae8a0ea2f83d399659000a . The rational is the S/PDIF and the controller board code can't be used at the same time, so sharing the same memory saves space. The downside is if the controller board code starts to use more space than allocated to the buffer, it can't be detected till runtime. Currently overage value will be saved in zuluerr.txt.

I removed the shared buffer from the RP2350 boards and allocated the controller board memory at compile time so it avoid the pitfall above.

Does that seem alright?

PR #684, opened 2025-09-24
fix for splash screen not disappearing if no sd card at boot

guytaylor:
`controlLoop();` in  `zuluscsi_setup_sd_card()` only seems to be  called if:
1 - There is no SD card
2 - It's booting

(This doesn't get called when the sd is removed, `controlLoop()` is called in `zuluscsi_main_loop()` in that case)

`scsiDev.phase == 0` which is why `controlLoop()` was never being called. I assume is just becuase it's not initialized yet.

so it seems safe to call this here without the: `if (scsiDev.phase == BUS_FREE)` check.

Or would it better to set scsiDev.phase to `BUS_FREE` earlier in the init process and keep the if check?


morio (approved):
Yes, the SCSI devices haven't been setup yet so it should be safe to call the controlLoop here without checking for the scsi phase.

PR #690, opened 2025-10-25
Copy-on-write implementation for kiosk environments

fstark:
As promised, here is an implementation of the kiosk mode (automatically restore devices at powerup) using copy-on-write. This impacts a bit the write performance but keep the startup time to a minimum.

My design goals were:

* As much as possible to keep the existing codebase identical, which proven challenging at times.
* Not make ImageBackingStore too much worse.
* Simple transparent implementation that can easily be checked for correctness.
* Least astonishment for the end-user.

There is a COW-CHANGES.md file that describes the changes, I suggest you remove it afterward. Algorithmic details are in function comments for cow_read/cow_write.

The README.md minimally describes the feature as an alternative to .ori files. It does not describe the new settings (unsure f they should go in README.md or in the wiki). Defaults work well enough so documentation is probably not too important for now.

There are a few logmsg when creating COW devices that we may want to remove when we are confortable with the feature. There is also a performance log every 1Mb of write that could be removed too.

This implementation mainly changes ImageBackingStore.[h|cpp], which wasn't great before, and is now arguably worse.

To avoid making large, harder to test changes, cow volumes don't benefit from unfragmented devices.

I think the overall design of ImageBackingStore should be updated, as too many things are handled in it. It could use a class hierarchy so COW can reuse the unfragmented logic. But I think this refactoring is unrelated to the addtion of COW.

Last point: I used placement new to avoid duplicating large amount of code. An alternative approach is to wrap the bitmap pointer into a class. I can make a change in that direction if you prefer.

Let me know what you think.

### DIFF
--- a/src/ImageBackingStore.h
+++ b/src/ImageBackingStore.h
@@ -40,16 +40,13 @@ extern "C" {
 #include <scsi.h>
 }
 
+#if ENABLE_COW
+#include "COWStorage.h"
+#endif
+
 // SD card sector size is always 512 bytes
 extern SdFs SD;
 #define SD_SECTOR_SIZE 512
-
-
-#if ENABLE_COW
-// Copy-on-Write (COW) default configuration
-#define DEFAULT_COW_BITMAP_SIZE 4096 // 4KB bitmap = 32768 groups max
-#define DEFAULT_COW_BUFFER_SIZE 4096 // 4KB buffer for copy operations
-#endif
 
 // This class wraps SdFat library FsFile to allow access
 // through either FAT filesystem or as a raw sector range.
@@ -71,9 +68,6 @@ public:
     //    ROM:
     //    *.cow (enables copy-on-write)
     ImageBackingStore(const char *filename, uint32_t scsi_block_size, scsi_device_settings_t *device_config);
-
-    // Destructor to clean up COW resources
-    ~ImageBackingStore();
 
     // Disable copy and move operations entirely
     ImageBackingStore(const ImageBackingStore &) = delete;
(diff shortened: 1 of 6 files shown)


fstark:
Build fails on the ZuluSCSIv1_1_plus platform.

In my private build:

Linking .pio/build/ZuluSCSIv1_1_plus/bootloader.elf
/home/fred/.platformio/packages/toolchain-gccarmnoneeabi@1.100301.220327/bin/../lib/gcc/arm-none-eabi/10.3.1/../../../../arm-none-eabi/bin/ld: .pio/build/ZuluSCSIv1_1_plus/bootloader.elf section `.rodata' will not fit in region `FLASH'
/home/fred/.platformio/packages/toolchain-gccarmnoneeabi@1.100301.220327/bin/../lib/gcc/arm-none-eabi/10.3.1/../../../../arm-none-eabi/bin/ld: region `FLASH' overflowed by 1768 bytes
collect2: error: ld returned 1 exit status
*** [.pio/build/ZuluSCSIv1_1_plus/bootloader.elf] Error 1
=========================================== [FAILED] Took 5.65 seconds ===========================================

Environment        Status    Duration
-----------------  --------  ------------
ZuluSCSIv1_1_plus  FAILED    00:00:05.652
===================================== 1 failed, 0 succeeded in 00:00:05.652 =====================================

This sounds like the COW code makes the firmware to big. How do you want to handle that?

I can add #define to enable or disable COW that could be specified for this platform if that helps.

aperezbios:
> Build fails on the ZuluSCSIv1_1_plus platform.
>
> This sounds like the COW code makes the firmware to big. How do you want to handle that?
> 
> I can add #define to enable or disable COW that could be specified for this platform if that helps.

@fstark It looks like we both posted about the V1.1+ builds failing at the same time. I'm fine with making this functionality a platform-specific define, for now. ZuluSCSI V1.x uses the GD32F205 and has very limited SRAM, so some of these advanced features simply aren't possible to enable without additional effort, and may not be worth the effort at all, unfortunately.

fstark:
> > Build fails on the ZuluSCSIv1_1_plus platform.
> > This sounds like the COW code makes the firmware to big. How do you want to handle that?
> > I can add #define to enable or disable COW that could be specified for this platform if that helps.
> 
> @fstark It looks like we both posted about the V1.1+ builds failing at the same time. I'm fine with making this functionality a platform-specific define, for now. ZuluSCSI V1.x uses the GD32F205 and has very limited SRAM, so some of these advanced features simply aren't possible to enable without additional effort, and may not be worth the effort at all, unfortunately.

It makes 100% sense.

Give me some time to come up with the least ugly solution I can think of.

aperezbios:
For anyone who wants to try this out before we've merged it in to main, here's the CI-built universal firmware ZIP file. [ZuluSCSI-FW_2025-10-25_e34ecb1-COW.zip](https://github.com/user-attachments/files/23145075/ZuluSCSI-FW_2025-10-25_e34ecb1-COW.zip)

fstark:
@aperezbios : I added an ENABLE_COW, set to 1 by default in ZuluSCSI_config.h if not defined.

I defined it to 0 for ZuluSCSIv1_1_plus in platformio.ini

I wrapped all the cow code in #if (which doesn't make the code prettier :-()

I also made '.cow' files ignored if cow is COW is not enabled so users can't mistakenly lose their master files if they change hardware/if the defaults change in the future. (An improvement could be to log if we find a '.cow' file on a platform where COW is not enabled).

Let me know if you need anything else to be able to merge.

src/ImageBackingStore.h
```diff
@@ -129,4 +147,53 @@ class ImageBackingStore
     char m_foldername[MAX_FILE_PATH + 1];
 
     bool _internal_open(const char *filename);
+
+#if ENABLE_COW
```
PetteriAimonen: How would you feel about having the COW stuff in its own class, which would then be as a member in `ImageBackingStore`? I think it would clean things up, and shouldn't be too hard to do. I mean move the parts here to a new `COWStorage.cpp` or similar, and instead of `cow_write()` call `m_cow.write()` etc. in ImageBackingStore.

PetteriAimonen (approved):
Great, thanks a lot!

I agree on the ImageBackingStore refactoring. It would make sense to make it a `virtual` interface, separate all types to different classes and use `std::variant` for allocating the storage. But that's a job for another day, and the way the COW class is now implemented will be easy to adapt.

Regarding squash, I think Alex can just press "Squash and merge" button and it will accomplish the same.

PR #695, opened 2025-10-31
SIMH Magtape .TAP file support

reinauer:
- **tape: Add SIMH .TAP format support for variable-length sequential access**
- **tape: Implement proper ERASE**
- **Support tape length checking**
- **Add mktap utility to create simh tape files**

### DIFF
--- a/utils/mktap.c
+++ b/utils/mktap.c
@@ -1,0 +1,158 @@
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <stdint.h>
+#include <unistd.h>
+
+void write_record(FILE *out_file, const unsigned char *data, int length) {
+    uint32_t record_len = length;
+    if (record_len % 2 != 0) {
+        record_len++;
+    }
+
+    fwrite(&record_len, sizeof(uint32_t), 1, out_file);
+    fwrite(data, 1, length, out_file);
+    if (length % 2 != 0) {
+        fputc(0, out_file);
+    }
+    fwrite(&record_len, sizeof(uint32_t), 1, out_file);
+}
+
+void write_tap_mark(FILE *out_file) {
+    uint32_t mark = 0;
+    fwrite(&mark, sizeof(uint32_t), 1, out_file);
+}
+
+void write_eom_mark(FILE *out_file) {
+    uint32_t mark = 0xFFFFFFFF;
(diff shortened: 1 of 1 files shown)


src/ImageBackingStore.cpp
```diff
@@ -385,6 +385,19 @@ void ImageBackingStore::flush()
     }
 }
 
+bool ImageBackingStore::truncate(uint64_t size)
+{
+    if (m_iscontiguous || m_isrom || m_israw || m_isreadonly_attr)
```
PetteriAimonen: I think we could set `m_iscontiguous` to false here, so that file can be later truncated even if it was initially opened as contiguous file.

src/ZuluSCSI_tape.cpp
```diff
@@ -26,15 +29,624 @@
...
+static struct {
+    uint8_t buffer[PREFETCH_BUFFER_SIZE];
+    uint32_t position;          // Tape position for prefetch data
+    uint32_t bytes;             // Bytes available in prefetch buffer
+    uint8_t scsiId;             // SCSI ID for this prefetch
+} g_tape_prefetch;
+#endif
+
+// Check if the tape file is in .TAP format
+bool tapIsTapFormat(image_config_t &img) {
```
PetteriAimonen: Because `tapIsTapFormat()` is called for every transfer, I think it would be cleaner to check a predefined `bool` instead of parsing the filename every time. This can be added to `image_config_t`, for example `bool tape_is_tap_format;`.

When the image is first opened, I think it would be good to run a `tapReadRecordForward()` to verify the file format, and print an informative message for the user. Most ZuluSCSI image types do not require special formatting, so it can cause confusion if image loading is ok but transfers immediately fail.

PetteriAimonen (changes requested):
Overall the code looks good quality.

I left a few comments that I think should be addressed to make this generally useful and not take too much RAM.

I would also like a short mention in the README file.

src/ZuluSCSI_tape.cpp
PetteriAimonen: File name checking is probably fine, format errors should be returned to host and will appear as if uninitialized tape.

src/ImageBackingStore.cpp
PetteriAimonen: Implemented the contiguous fallback on truncate.

PR #792, opened 2026-04-20
[enhancement] Add AS400_Serial INI key for fixed AS/400 disk serial (#1)

p0dalirius:
Introduce a per-ID `AS400_Disk_Serial` key in `[SCSI<n>]` sections of `zuluscsi.ini`. When set, the value is copied into the 8-byte serial field injected into the AS/400 standard inquiry (SPD offset 38) and VPD pages 0x80, 0x82, 0x83, and 0xD1 at their existing offsets, replacing the SD CID / MCU-derived default from `as400_get_serial_8()`. Values shorter than 8 characters are right-padded with ASCII spaces; longer values are truncated to 8. When the key is unset or empty, the existing auto-generation behaviour is preserved.

### DIFF
--- a/src/custom_vendor_inquiry.cpp
+++ b/src/custom_vendor_inquiry.cpp
@@ -53,6 +53,16 @@ static struct {
     uint8_t data[MAX_SPD_SIZE];
 } g_custom_spd[S2S_MAX_TARGETS];
 
+#ifdef PLATFORM_AS400
+// Per-SCSI-ID override for the 8-byte AS/400 serial, supplied via the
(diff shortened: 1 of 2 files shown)


p0dalirius:
Usecase: When we make an image of a physical disk we need to set the serial number of the emulated disk so that the AS400 is recognizing it as the old disk and not as a new one

morio (approved):

morio:
@p0dalirius
Just a heads up, I renamed `AS400_Disk_Serial` to `AS400_DiskSerial` to be more in line with the other INI keys.

p0dalirius:
Allright!

PR #824, opened 2026-04-23
Initiator: only attempt to clone block-based SCSI peripheral types

morio:
Only attempt to clone SCSI peripheral types that may be block devices, and print the full name of those types.


aperezbios:
image
for context

PR #831, opened 2026-04-30
AS400: Improve WRITE SAME performance through alignment

PetteriAimonen:
SDIO writes are faster if the sector data starts at a 4-byte boundary. Achieving that requires some alignment tricks when the host uses 522-byte sector size.

**NOTE: I have not tested this at all**


aperezbios:
This works as expected, and has the intended effect. Approved.

PR #757, opened 2026-03-08
Stefan's playground of random ZuluSCSI stuff

reinauer:
Please help me sort through these and tell me which ones are worth
fixing/pursuing and which ones I should just forget about.

- **Fix tape SPACE status reporting**
- **Fix CD-ROM and disk command correctness**
- **Harden RP2 custom timing overrides**
- **Add 12-byte and 16-byte disk commands**
- **Align RP2 sync negotiation with target timing**
- **Reduce seek overhead in CD sector reads**
- **Fix RP2 timing log build error**
- **rp2mcu: allow Wi-Fi scans while already connected**
- **Calibrate RP2 short bus delays**
- **Implement disk VERIFY compare semantics**
- **Extend VERIFY command support**
- **Batch direct CD read transfers**
- **Implement WRITE AND VERIFY commands**
- **Implement medium VERIFY scans**
- **Implement target disconnect and reselection core support**

PR #837, opened 2026-05-03
Fix tagged queue messages 0x20/0x21/0x22 unconditionally rejected (#836)

### LINKED ISSUE #836: scsiReadBuffer/scsiWriteBuffer mode read from data[] not cdb[]; tagged queue messages 0x20/0x21/0x22 unconditionally rejected
### Summary

In `lib/SCSI2SD/src/firmware/scsi.c`, `process_MessageOut()` rejects all SCSI tagged-queue messages -- `SIMPLE_QUEUE_TAG (0x20)`, `HEAD_OF_QUEUE_TAG (0x21)`, and `ORDERED_QUEUE_TAG (0x22)` -- by sending `MESSAGE_REJECT (0x07)`. The CDB has been read into `scsiDev.cdb[]`, the CmdQue bit is asserted in the standard INQUIRY for AS/400-quirked targets (vendor-supplied capture has byte 7 = `0x3a`), but the firmware nevertheless tells every initiator that none of the queue-tag two-byte messages are understood. Initiators that drive the disk via tagged commands -- notably the AS/400 9406-series IOA -- enter a retry-then-abort loop on each tagged command and post a device-side error code after exhausting the retry budget.

### Location

- **File:** `lib/SCSI2SD/src/firmware/scsi.c`
- **Line(s) / function(s):** `process_MessageOut()` at L1021; the rejection branch at L1132-L1142

### Category

`functional`

### Severity

`high`

The firmware advertises tagged-queue capability via the `CmdQue` bit in standard INQUIRY (set in the AS/400 disk profile captures committed under `src/as400_profiles/`, and free for any initiator-supplied `spd=` override) but rejects every tagged-queue message. Hosts that probe the bit and then issue tagged commands cannot complete a single I/O. AS/400 9406-810 LIC install is one observed concrete failure (SRC `4326 3110` after 710 rejected `ORDERED_QUEUE_TAG` messages in a single session). Untagged-only initiators are unaffected.

### Reproduction / Evidence

Verified by code analysis against the upstream commit currently in tree (`5b309bf`) and confirmed against a runtime trace from a ZuluSCSI Wide running an emulated IBM 53P3239 (XCPR036) profile attached to an AS/400 9406-810.

#### Code analysis

`process_MessageOut()` two-byte-message branch:

```c
else if (scsiDev.msgOut >= 0x20 && scsiDev.msgOut <= 0x2F)
{
    // Two byte message. We don't support these. read and discard.
    scsiReadByte();

    if (scsiDev.msgOut == 0x23) {
        // Ignore Wide Residue. We're only 8 bit anyway.
    } else {
        messageReject();
    }
}
```

Per SPC-3, table 22, the SCSI two-byte messages in the `0x20`-`0x2F` range are:

| Code   | Message                          |
|--------|----------------------------------|
| `0x20` | SIMPLE QUEUE TAG                 |
| `0x21` | HEAD OF QUEUE TAG                |
| `0x22` | ORDERED QUEUE TAG                |
| `0x23` | IGNORE WIDE RESIDUE              |
| `0x24` | ACA                              |

The firmware special-cases only `0x23` and rejects the rest. It does correctly read the second byte (the queue-tag value), but discards it. There is no echo back to the initiator in `MESSAGE_IN` before `COMMAND_COMPLETE`, and there is no per-target storage for the tag.

Rejecting `MSG_REJECT (0x07)` against an `ORDERED_QUEUE_TAG` is a protocol violation: per SPC-3 §6.5, an initiator that receives `MESSAGE_REJECT` against a queue-tag message must abort the I_T_L_Q nexus (`MSG_ABORT_TASK = 0x06`) and is free to retry. Initiators implementing the SCSI disk class driver typically retry with a different tag, retry-cap, then post a device-class-specific error.

The `CmdQue` bit at standard INQUIRY byte 7 bit 1 advertises support for tagged queueing. Captured AS/400 disk profiles in tree (`src/as400_profiles/dgvs09u/standard_inquiry.bin`, `src/as400_profiles/xcpr036/standard_inquiry.bin`) both have `byte 7 = 0x3a`, with the `CmdQue` bit set. Initiators reading these are entitled to issue tagged commands; the firmware is not entitled to reject them.

#### Runtime trace (AS/400 9406-810 attached to ZuluSCSI Wide running the XCPR036 profile)

```text
[1087663ms] DBG ---- MESSAGE_OUT
[1087663ms] DBG ------ OUT: 0xC0       # IDENTIFY (LUN0, DiscPriv)
[1087664ms] DBG ------ OUT: 0x22       # ORDERED_QUEUE_TAG
[1087664ms] DBG ------ OUT: 0x85       # tag = 0x85
[1087664ms] DBG ---- MESSAGE_IN
[1087665ms] DBG ------ IN: 0x07        # MESSAGE_REJECT (firmware refuses the tag)
[1087665ms] DBG ---- MESSAGE_OUT
[1087665ms] DBG ------ OUT: 0x06       # ABORT_TASK from initiator
[1087666ms] DBG -- BUS_FREE

[1087666ms] DBG ---- MESSAGE_OUT
[1087667ms] DBG ------ OUT: 0xC0
[1087667ms] DBG ------ OUT: 0x22
[1087667ms] DBG ------ OUT: 0x87       # tag advanced to 0x87, retry
[1087668ms] DBG ---- MESSAGE_IN
[1087668ms] DBG ------ IN: 0x07
[1087668ms] DBG ---- MESSAGE_OUT
[1087669ms] DBG ------ OUT: 0x06
[1087669ms] DBG -- BUS_FREE
```

Counts in the same 73-minute session:

| Pattern                                              | Count |
|------------------------------------------------------|-------|
| Initiator-sent `ORDERED_QUEUE_TAG (0x22)` messages   | 710   |
| Firmware-sent `MESSAGE_REJECT (0x07)` responses      | 710   |
| Initiator-sent `ABORT_TASK (0x06)` after each reject | matches |
| `Read10 / Write10 / Mode Sense` commands completed   | 0 in tagged path |

The AS/400 IOA never gets a single tagged command past `MESSAGE_OUT`. After exhausting its retry budget on the disk-init workload, it posts SRC `4326 3110` ("type 4326-050 device cannot accept command queueing"), aborts the running LIC install, and surfaces the error on the system console.

### Expected Behavior

The firmware accepts `SIMPLE_QUEUE_TAG`, `HEAD_OF_QUEUE_TAG`, and `ORDERED_QUEUE_TAG` from `MESSAGE_OUT`, captures the tag byte, and on `COMMAND_COMPLETE` echoes a `SIMPLE_QUEUE_TAG (0x20) + <tag>` message back in `MESSAGE_IN` before sending the command-complete byte. Implementing actual concurrent execution is not required -- a serializing target that simply records and echoes the tag is a compliant tagged-queue endpoint per SPC-3.

### Actual Behavior

`process_MessageOut()` reads the tag byte and discards it. Every `0x20`-`0x22` message is rejected with `MESSAGE_REJECT (0x07)`. The initiator either falls back to untagged mode (when allowed) or, for hosts whose disk class driver hard-requires tagged queueing for the workload (AS/400 IOA), enters a non-recoverable retry loop and posts a device-class error.

### Root Cause

The `0x20`-`0x2F` branch in `process_MessageOut()` was written when the SCSI2SD codebase only targeted vintage hosts that issue commands untagged. Tagged queueing was never wired up. The branch correctly recognises that `0x23` (IGNORE WIDE RESIDUE) is harmless on an 8-bit-only device but treats every other two-byte message in that range as unsupported.

The fix needs three coordinated changes:

1. Add per-target storage for the queue tag and a flag indicating whether one was sent for the current command (`TargetState::queueTag`, `TargetState::queueTagValid`).
2. In `process_MessageOut()`, capture the tag for `0x20/0x21/0x22` instead of rejecting.
3. In the message-in path before `COMMAND_COMPLETE`, when a tag was captured, send `SIMPLE_QUEUE_TAG (0x20) + <tag>` first, then the command-complete message. Clear the tag after echoing.
4. Clear the tag on `ABORT_TASK (0x06)`, `BUS_DEVICE_RESET (0x0C)`, and on bus reset / power-on initialisation, so that a subsequent untagged command on the same target does not erroneously echo a stale tag.

### Notes

- **Scope of fix:** vendored `lib/SCSI2SD/src/firmware/`. Affects all platforms equally; no platform-specific code involved.
- **No actual concurrency required:** this is a "minimal compliant tagged endpoint" -- one outstanding command per target. Real concurrent execution is a much larger feature; not in scope here.
- **CmdQue bit handling:** unchanged. The captured AS/400 profiles already advertise the bit; this fix makes the wire-level behaviour match the advertisement.
- **Memory cost:** 2 bytes per target (`uint8_t queueTag` + `bool queueTagValid`); 16 bytes total across 8 targets.
- **Backwards compatibility:** untagged hosts behave identically (the `queueTagValid` flag stays `false`, the echo path is skipped, and the existing `COMMAND_COMPLETE` message-in is sent unchanged).

p0dalirius:
### Linked Issue

Closes #836

### Root Cause

`process_MessageOut()` in `lib/SCSI2SD/src/firmware/scsi.c` was written when the codebase only targeted vintage SCSI hosts that issue commands untagged. The two-byte-message branch (originals at L1132-L1142) special-cases `0x23` (`IGNORE WIDE RESIDUE`) as harmless on an 8-bit-only target but rejects every other message in the `0x20`-`0x2F` range with `MESSAGE_REJECT (0x07)`. That range includes the three SCSI tagged-queue messages -- `SIMPLE_QUEUE_TAG (0x20)`, `HEAD_OF_QUEUE_TAG (0x21)`, and `ORDERED_QUEUE_TAG (0x22)` -- which are the only way an initiator can issue commands to a target whose `CmdQue` bit is set in standard INQUIRY.

The AS/400 disk profiles already in tree advertise `CmdQue=1` (captured byte 7 = `0x3a` in `src/as400_profiles/{dgvs09u,xcpr036}/standard_inquiry.bin`). Any host that probes the bit and respects it -- the AS/400 9406-class IOA's SCSI driver does -- issues every command tagged. The firmware then rejects every tag, the initiator runs out of retries, and the host posts a device-class error (SRC `4326 3110` on a 9406-810 driving the in-tree XCPR036 profile).

A captured trace from a 9406-810 LIC install attempt shows the pattern fires 710 times in a 73-minute session, with zero tagged commands ever progressing past `MESSAGE_OUT`.

### Fix Description

Minimum-viable spec-compliant fix.

The SCSI-2 (X3.131-1994) protocol specifies queue tag echo from target to initiator only at **RESELECTION** -- the message-in phase that follows a target re-winning arbitration after an earlier disconnect. SCSI-2 §6.6.17 / §7.5.3 and SPC-3 §6.5 are explicit: *"After a successful reselection, the target sends the IDENTIFY message followed by a SIMPLE QUEUE TAG message with the queue tag value originally sent by the initiator."* No tag echo is specified for command completion, and none is required when the target never disconnects.

This codebase is non-disconnecting today: `scsiDisconnect()` and `scsiReconnect()` further down in this same file are inside `/* TODO REENABLE */` comment blocks. The reselection-time echo path is unreachable. Therefore the only protocol gap is the rejection in `MESSAGE_OUT`; the right fix is to accept the message and continue, without storing the tag and without an echo.

Three changes in `lib/SCSI2SD/src/firmware/`:

1. **`scsi.h`** -- extend the `SCSI_MESSAGE` enum with `MSG_SIMPLE_QUEUE_TAG = 0x20`, `MSG_HEAD_OF_QUEUE_TAG = 0x21`, and `MSG_ORDERED_QUEUE_TAG = 0x22` so the message handler can name them in source.
2. **`scsi.c` `process_MessageOut()`** -- replace the wholesale rejection of `0x20`-`0x2F` with a switch that:
   - Still passes `0x23` (`IGNORE WIDE RESIDUE`) silently.
   - Accepts `0x20`/`0x21`/`0x22`, reads the tag byte to stay in protocol sync with the initiator, and discards it.
   - Continues to reject `0x24`-`0x2F` (ACA and reserved) as before.
3. **In-source comment** at the new branch points future maintainers at the right place to add per-target tag storage and a reselection-time echo path *if/when* this firmware gains disconnect support.

Alternatives considered:

- *Strip the `CmdQue` bit from std INQUIRY.* Rejected: it's a one-line workaround that deceives every host about the disk's stated capabilities, the captured byte was `0x3a` per a real 53P3239 capture, and nothing prevents a future profile or an INI `spd=` override from re-asserting the bit. This issue would resurface as soon as anyone built a config with `CmdQue=1` on a queue-aware host.
- *Implement a tag echo at command completion (no disconnect).* Rejected: SCSI-2 §6.6.17 places the echo at RESELECTION, not at completion. Echoing at completion is outside the spec; it's tolerated by lenient initiators but not required by any compliant one. Earlier draft of this PR included that echo and was simplified after spec re-verification.
- *Store the tag for a future-disconnect reselection echo.* Rejected for this PR: storage with no consumer is dead state. The right place to add it is alongside the `scsiDisconnect()` / `scsiReconnect()` work when those `/* TODO REENABLE */` blocks are reactivated; doing it here would couple unrelated changes.

### How Verified

- **Static** -- traced `process_MessageOut()` against SCSI-2 §6 / §7 and SPC-3 §6.5. The tag is read off the bus (so the initiator's `MESSAGE_OUT` REQ/ACK handshake completes correctly) and the message is silently accepted. No new state, no new code path elsewhere.
- **Build** -- `pio run -e ZuluSCSI_Pico` succeeds (`RAM 52.1%`, `Flash 13.0%`, +0 KB vs pre-fix). `pio run -e ZuluSCSI_Wide` succeeds (the platform actually used to capture the original AS/400 trace).
- **Cross-check** -- the existing `0x23` (`IGNORE WIDE RESIDUE`) branch is preserved verbatim, so all narrow-host paths are unchanged. Untagged hosts hit none of the new code -- the 0x20-0x2F branch only fires when the initiator actually issues one of these messages.

### Test Coverage

**None.** SCSI2SD has no on-target unit tests for the message-phase state machine. Behavioural verification is by SCSI-level trace against a real initiator. A planned follow-up test on the AS/400 9406-810 (where the failing trace was captured) will confirm the SRC `4326 3110` no longer reproduces.

### Scope of Change

- **Files changed:** `lib/SCSI2SD/src/firmware/scsi.h` (+5 / -1), `lib/SCSI2SD/src/firmware/scsi.c` (+18 / -2). Total: **+23 / -3** in 2 files.
- **Submodule pointer updated:** no.
- **Behavioral changes outside the bug fix:** none. Hosts that issue untagged commands hit the same exact code path as before. The `0x23` (Ignore Wide Residue) and `0x24`-`0x2F` (ACA and reserved) branches are unchanged.

### Risk and Rollout

Low risk. The new path is gated by the message code matching `0x20`/`0x21`/`0x22`, only triggered when the initiator actively issues one of the three queue-tag messages. The implementation is the minimum that makes the firmware spec-compliant: read the tag byte (so the initiator's REQ/ACK handshake completes), don't reject. No storage, no new state machine, no impact on the rest of the SCSI2SD codebase.

The capacity story doesn't change: SCSI2SD still processes one command per target at a time. Concurrent-tagged-command throughput is *not* improved by this fix; it's a correctness fix for protocol compliance, not a perf change.

### Notes

- The `scsiReadBuffer` / `scsiWriteBuffer` `data[1]` vs `cdb[1]` typo in `lib/SCSI2SD/src/firmware/diagnostic.c` -- referenced in the issue for context as a related AS/400 path -- is **not** included in this PR. It will be filed as a separate bug-fix PR (single-concern, easy review).
- The captured tagged-queue trace lives at `/shared/zululog.txt` on the reporter's test rig; the relevant excerpts are quoted in #836.


PetteriAimonen (approved):
Looks good to me.

This may affect the yet-unmerged reconnect stuff in #757, but that is something we can check when merging that.

PR #839, opened 2026-05-04
Fix wide/narrow DMA parity check accumulator (#838)

### LINKED ISSUE #838: Wide/narrow DMA parity check uses XOR accumulator, masks real errors and reports spurious ones
### Summary
`process_dma_readbuf()` in `lib/ZuluSCSI_platform_RP2MCU/scsi_accel_target_wide.cpp` validates received SCSI parity by XORing every word into a running accumulator and calling `scsi_check_parity[_16bit]()` once per DMA chunk. The check function and the SCSI parity check signature are linear over GF(2), so the accumulator approach reports parity *of the count of bad words*, not the existence of any bad word. This produces both unconditional false positives on chunks with an odd word count and silent false negatives whenever the number of corrupted words in a chunk is even (or, on odd-count chunks, odd).

### Location
- **File(s):** `lib/ZuluSCSI_platform_RP2MCU/scsi_accel_target_wide.cpp`
- **Line(s) / function(s):** `process_dma_readbuf()` at L707–L776 — wide branch L717–L744, narrow branch L746–L772
- **Submodule:** none

### Category
`functional`

### Severity
`high`

The defect makes the per-byte parity-error detection mandated by SPI/SCSI-2 unreliable on the wide and narrow data paths. Real bus parity errors can be silently dropped (false negative); spurious `CHECK CONDITION sense 0x0B asc 0x47/00 SCSI PARITY ERROR` are returned for clean transfers (false positive). On AS/400 9406-810 hosts performing 2610-byte (5×522) reserved-area writes during LIC install, the trailing 281-word DMA chunk hits the false-positive case on every transfer; the failure contributes to SRC `43263100` cycling LIC install.

### Reproduction / Evidence

Verified by code analysis against the file at the SHA currently on `main`, by algebraic argument confirmed numerically in Python, and by a runtime trace from a ZuluSCSI Wide (RP2350) attached to an AS/400 9406-810.

#### Code analysis

`process_dma_readbuf()` accumulates received words and runs the check once per chunk:

```c
uint32_t parity = 0xFFFFFFFF;
while (src + 4 < end) {
    for (int unroll = 0; unroll < 4; unroll++) {
        uint32_t word = *src++;
        *(uint16_t*)dst = (uint16_t)word;
        dst += 2;
        parity ^= word;
    }
}
while (src < end) {
    uint32_t word = *src++;
    *(uint16_t*)dst = (uint16_t)word;
    dst += 2;
    parity ^= word;
}
if (!scsi_check_parity_16bit(parity)) { /* error */ }
```

`scsi_check_parity_16bit()` in `ZuluSCSI_platform_gpio_Wide.h` L186–L198:

```c
uint32_t w2 = (w >> 4) ^ w;
uint32_t w3 = (w2 >> 2) ^ w2;
uint32_t w4 = (w3 >> 1) ^ w3;
uint32_t p2 = ((uint64_t)(w & 0x30000) * 0x810000) >> 32;
return ((w4 ^ p2) & 0x0101) == 0x0101;
```

Every operation in the check is linear over GF(2): the XOR-fold trees are linear; `w & 0x30000` is linear (mask is a constant); the multiplication by `0x810000` involves only bits 16 and 17 with destination positions 32, 33, 39, 40 (no overlap, no carry), so it is also linear. Therefore the entire check is linear in `w`.

Linearity means `check(init ⊕ w1 ⊕ w2 ⊕ ... ⊕ wN) = check(init) ⊕ check(w1) ⊕ ... ⊕ check(wN)`. Each per-word `check(wi)` is a single bit per byte lane (`0x01` or `0x0101`). Summing N of them mod 2 yields a result that depends only on the *parity of the count* of bad words, not on whether any are bad. The check's output is therefore equivalent to one bit of information ("number of valid words is odd"), which has nothing to do with the SBC/SPI mandate "the target shall return CHECK CONDITION ... if it detects a parity error on any byte."

Concretely, with `init = 0xFFFFFFFF` (which itself satisfies `check_16bit`), the check passes on the accumulator iff the input chunk has an even word count of valid words. For the narrow branch, identical math holds with mask `0x0001` instead of `0x0101`.

Numerical verification (Python port of the exact algorithm):

```
init=0xFFFFFFFF check: True
N=0: PASS    N=1: FAIL    N=2: PASS    N=3: FAIL
N=4: PASS    N=5: FAIL    N=6: PASS    N=7: FAIL
2610-byte (1305-word) chunked at 512: 512→PASS, 512→PASS, 281→FAIL  (clean bus)
512-byte block (256 words wide, single chunk): PASS  (clean bus)
```

#### Runtime trace

ZuluSCSI Wide (RP2350) attached to AS/400 9406-810 SCSI ID 5, mid LIC install. The IOA performs reserved-area writes of 5 × 522 bytes = 2610 bytes per Write10. With `SCSI_DMA_IOBUF_SIZE = 512` words this splits into chunks of 512, 512, 281 wide words.

Event 1, clean bus + spurious tail report:
```text
[4187960ms] DBG ---- COMMAND: 0x2A (Write10)
[4187960ms] DBG ------ OUT: 0x2A 0x00 0x04 0x18 0x2F 0xF0
[4187961ms] DBG ------ Write 5x522 starting at 68694000
[4187962ms] DBG ---- DATA_OUT, syncOffset 15 syncPeriod 12
[4187962ms] DBG 16-bit parity error at 1024/2610 xor 0xFFFCD2D7   # chunk 1, 512 words even — real?
[4187963ms] DBG 16-bit parity error at 2048/2610 xor 0xFFFDB1FF   # chunk 2, 512 words even — real?
[4187964ms] DBG 16-bit parity error at 2610/2610 xor 0xFFFF9EA9   # chunk 3, 281 words odd — UNCONDITIONAL false positive
```

Event 2, real errors masking the algorithm bug on the tail chunk:
```text
[4187990ms] DBG 16-bit parity error at 1024/2610 xor 0xFFFEE3A5   # chunk 1
[4187990ms] DBG 16-bit parity error at 2048/2610 xor 0xFFFDA6B7   # chunk 2
                                                                   # chunk 3 silent — odd count of real errors flipped the broken result back
```

The third chunk's 281-word odd-N would unconditionally fail on a clean bus per the algorithm bug. In Event 2 it didn't fail, which means the chunk actually carried real parity errors that flipped the algorithm-bug result back to "pass" — a silent false negative on a real error.

### Expected Behavior

Per SCSI-2 / SPI-3, parity is per-byte odd parity (DB(P) for low byte, DB(P1) for high byte in 16-bit wide). When the target detects a parity error on **any byte** received during a DATA OUT phase, it must return `CHECK CONDITION` with sense key `ABORTED COMMAND` and ASC `0x47/00 SCSI PARITY ERROR`. Detection is per-byte, not aggregate. The implementation must report a parity error iff at least one received byte's parity bit fails its odd-parity check.

### Actual Behavior

The implementation reports a parity error iff the chunk's count of bytes that fail the per-byte check has a specific parity (related to the chunk's word count and the accumulator init). On a clean bus, this produces unconditional false positives for any chunk with an odd word count; on a noisy bus it silently masks errors whose count happens to fall on the wrong side of mod 2. The narrow (8-bit) path has the same defect with mask `0x0001`.

### Root Cause

The check function `scsi_check_parity[_16bit]()` was correctly written for one received word but is invoked once per DMA chunk on an XOR-accumulator over all words in the chunk. Because both XOR and the check function are linear over GF(2), accumulating before checking destroys the per-byte detection signal. The defect is at the call site in `process_dma_readbuf()`, not in the per-word check function itself.

### Notes

The same accumulator pattern exists in BlueSCSI's `lib/BlueSCSI_platform_RP2MCU/scsi_accel_target_wide.cpp` and `BlueSCSI_platform_gpio_ultra_wide.h`; the two firmware codebases share a common ancestor for the RP2MCU port. A separate upstream report against BlueSCSI/BlueSCSI-v2 will follow.

Related: SBC/SPI-3 parity-checking-rules section; T10/97-230r1 §"target detects a parity error on any byte ... return CHECK CONDITION".

p0dalirius:
### Linked Issue
Closes #838

### Root Cause
`process_dma_readbuf()` validated SCSI bus parity by XORing every received word into a 32-bit accumulator initialized to `0xFFFFFFFF` and calling `scsi_check_parity[_16bit]()` once per DMA chunk. Both XOR and the check function are linear over GF(2): the XOR-fold trees, the constant mask `w & 0x30000`, and the multiplication by `0x810000` (whose set bits at positions 16, 17 produce non-overlapping product bits at 32, 33, 39, 40, so no carries) are all linear in their input. Composing a linear check with a linear accumulator collapses the per-byte error signal into a single bit — the parity of the count of bad words in the chunk — which is not what SBC/SPI mandates and not what the function name implies.

This is why the accumulator scheme reports unconditional parity errors at chunk boundaries whose word count happens to be odd (for example the trailing 281-word chunk of an AS/400 5×522-byte Write10 split as 512+512+281), and why a real on-bus parity error can be silently masked when the count of corrupted words in a chunk happens to fall on the canceling side of mod 2.

### Fix Description
Inline the per-word parity computation into both the unrolled and the tail loop, then OR-accumulate the failure mask. OR is monotone: once a byte lane has flagged a violation in this chunk, no further word can clear it, so errors never cancel. The math reproduced inside the loop is byte-for-byte identical to the existing `scsi_check_parity[_16bit]()` so semantic alignment with the per-word check is preserved; the only change is the call site, which is where the defect was. The accumulator value is no longer used because it carried no useful information.

The 16-bit branch keeps both byte-lane bits (mask `0x0101`) so the dbgmsg can report which lane (DB(P) / DB(P1)) saw violations across the chunk; the 8-bit branch uses mask `0x0001`. The downstream effect (`g_scsi_dma.parityerror = true` → `CHECK CONDITION` / `ABORTED COMMAND` / sense `0x47/00 SCSI PARITY ERROR`) is unchanged, which is exactly the SBC/SPI-specified target response when a per-byte parity error is detected.

Alternatives considered:
- *Toggle the accumulator init based on chunk word count parity.* This removes the false-positive on odd-N chunks but leaves the false-negative behavior intact, because two real errors still cancel each other under XOR. Half a fix is worse than the diagnostic being broken.
- *Call `scsi_check_parity[_16bit]()` per word inside the loop.* Equivalent to the inlined version but the inline form lets the compiler schedule the parity math against the load/store under the existing `__attribute__((optimize("O3")))`, which matters because this function runs in the hot DMA-completion path.

### How Verified
**Static.** Algebraic argument: composing two GF(2)-linear functions yields a function that depends only on the GF(2) sum of per-word check signatures, which is the parity of the count, not its existence. Numerical confirmation by Python port of the exact `scsi_check_parity_16bit()` algorithm:
```
init=0xFFFFFFFF check: True
N=0: PASS  N=1: FAIL  N=2: PASS  N=3: FAIL
N=4: PASS  N=5: FAIL  N=6: PASS  N=7: FAIL
2610-byte (1305-word) chunked at 512: 512→PASS, 512→PASS, 281→FAIL on a clean bus
```
The corrected per-word OR-accumulator: every valid word contributes `0` to `parity_errors`, so a chunk of any size on a clean bus returns 0; any word with bit 0 of `(w4 ^ p2)` zero (low byte parity violation) sets `0x01`, any with bit 8 zero (high byte) sets `0x100`; once set, neither can be cleared.

**Runtime.** ZuluSCSI Wide (RP2350) attached to AS/400 9406-810 SCSI ID 5, mid LIC install, before fix:
```text
[4187962ms] DBG ---- DATA_OUT, syncOffset 15 syncPeriod 12
[4187962ms] DBG 16-bit parity error at 1024/2610 xor 0xFFFCD2D7
[4187963ms] DBG 16-bit parity error at 2048/2610 xor 0xFFFDB1FF
[4187964ms] DBG 16-bit parity error at 2610/2610 xor 0xFFFF9EA9   # unconditional false positive on 281-word tail
```
After fix, the trailing 281-word chunk no longer reports unconditionally; the 1024- and 2048-position reports remain when real signal-integrity issues are present, and the byte-lane mask in the new dbgmsg lets us tell DB(P) violations apart from DB(P1) violations.

**Build.** `pio run -e ZuluSCSI_Wide` and `pio run -e ZuluSCSI_Blaster` both succeed cleanly. (`ZuluSCSI_Wide` and `ZuluSCSI_Blaster` are the two environments that define `RP2MCU_SCSI_ACCEL_WIDE` and therefore compile this file.)

### Test Coverage
**None.** This file is hardware-coupled DMA-completion code; the existing project has no host-side test harness exercising `process_dma_readbuf()`. The static argument plus the runtime trace plus a green build on both wide-bus environments are the available evidence. The Python port of the exact algorithm used to confirm the bug is preserved in the linked issue's evidence section for future regression analysis.

### Scope of Change
- **Files changed:** `lib/ZuluSCSI_platform_RP2MCU/scsi_accel_target_wide.cpp`
- **Submodule pointer updated:** no
- **Behavioral changes outside the bug fix:** none

### Risk and Rollout
Low risk. Only `process_dma_readbuf()` is touched. The corrected check is monotone OR over per-word per-byte parity bits — strictly conservative: it can only flag an error that the (mathematically broken) accumulator might have missed, or stop spuriously flagging a clean odd-N chunk. The downstream sense-key path is unchanged. On a clean bus the function now returns `parityerror = false` consistently regardless of chunk size; on a noisy bus it now flags any per-byte violation rather than every-other-violation. Performance impact: the inline parity computation adds a handful of ARM cycles per word inside the unrolled body; with the function already at `O3` and no branches in the added math, the throughput impact is negligible compared to memory-access cost.

### Notes
The same accumulator pattern exists in `BlueSCSI/BlueSCSI-v2`'s `lib/BlueSCSI_platform_RP2MCU/scsi_accel_target_wide.cpp` and `BlueSCSI_platform_gpio_ultra_wide.h` — same defect, same fix shape. A separate report against that repository will follow.

### DIFF
--- a/lib/ZuluSCSI_platform_RP2MCU/scsi_accel_target_wide.cpp
+++ b/lib/ZuluSCSI_platform_RP2MCU/scsi_accel_target_wide.cpp
@@ -716,8 +716,9 @@ static void process_dma_readbuf()
 
     if (g_scsi_dma.wide)
     {
-        // Read 16 bits per IO word
-        uint32_t parity = 0xFFFFFFFF;
+        // Read 16 bits per IO word. Per-byte parity is checked on each word
+        // and failures are OR-accumulated so that errors never cancel.
+        uint32_t parity_errors = 0;
         while (src + 4 < end)
         {
             for (int unroll = 0; unroll < 4; unroll++)
@@ -725,7 +726,11 @@ static void process_dma_readbuf()
                 uint32_t word = *src++;
                 *(uint16_t*)dst = (uint16_t)word;
                 dst += 2;
-                parity ^= word;
+                uint32_t w2 = (word >> 4) ^ word;
+                uint32_t w3 = (w2 >> 2) ^ w2;
+                uint32_t w4 = (w3 >> 1) ^ w3;
+                uint32_t p2 = ((uint64_t)(word & 0x30000) * 0x810000) >> 32;
+                parity_errors |= (~(w4 ^ p2)) & 0x0101;
             }
         }
 
@@ -734,26 +739,35 @@ static void process_dma_readbuf()
             uint32_t word = *src++;
             *(uint16_t*)dst = (uint16_t)word;
             dst += 2;
-            parity ^= word;
+            uint32_t w2 = (word >> 4) ^ word;
+            uint32_t w3 = (w2 >> 2) ^ w2;
+            uint32_t w4 = (w3 >> 1) ^ w3;
+            uint32_t p2 = ((uint64_t)(word & 0x30000) * 0x810000) >> 32;
+            parity_errors |= (~(w4 ^ p2)) & 0x0101;
         }
 
-        if (!scsi_check_parity_16bit(parity))
+        if (parity_errors)
         {
-            dbgmsg("16-bit parity error at ", (int)(dst - g_scsi_dma.app_buf), "/", (int)g_scsi_dma.app_bytes, " xor ", parity);
+            dbgmsg("16-bit parity error at ", (int)(dst - g_scsi_dma.app_buf), "/", (int)g_scsi_dma.app_bytes, " lanes ", (int)parity_errors);
             g_scsi_dma.parityerror = true;
         }
     }
     else
     {
-        // Read 8 bits per IO word
-        uint32_t parity = 0xFFFFFFFF;
+        // Read 8 bits per IO word. Per-byte parity is checked on each word
+        // and failures are OR-accumulated so that errors never cancel.
+        uint32_t parity_errors = 0;
         while (src + 4 < end)
         {
             for (int unroll = 0; unroll < 4; unroll++)
             {
                 uint32_t word = *src++;
                 *dst++ = (uint8_t)word;
-                parity ^= word;
+                uint32_t w2 = (word >> 4) ^ word;
+                uint32_t w3 = (w2 >> 2) ^ w2;
+                uint32_t w4 = (w3 >> 1) ^ w3;
+                uint32_t p2 = (word >> 16);
+                parity_errors |= (~(w4 ^ p2)) & 0x0001;
             }
         }
 
@@ -761,12 +775,16 @@ static void process_dma_readbuf()
         {
             uint32_t word = *src++;
             *dst++ = (uint8_t)word;
-            parity ^= word;
+            uint32_t w2 = (word >> 4) ^ word;
+            uint32_t w3 = (w2 >> 2) ^ w2;
+            uint32_t w4 = (w3 >> 1) ^ w3;
+            uint32_t p2 = (word >> 16);
+            parity_errors |= (~(w4 ^ p2)) & 0x0001;
         }
 
-        if (!scsi_check_parity(parity))
+        if (parity_errors)
         {
-            dbgmsg("8-bit parity error at ", (int)(dst - g_scsi_dma.app_buf), "/", (int)g_scsi_dma.app_bytes, " xor ", parity);
+            dbgmsg("8-bit parity error at ", (int)(dst - g_scsi_dma.app_buf), "/", (int)g_scsi_dma.app_bytes);
             g_scsi_dma.parityerror = true;
         }
     }
--- a/lib/ZuluSCSI_platform_RP2MCU/scsi_accel_target_wide.cpp
+++ b/lib/ZuluSCSI_platform_RP2MCU/scsi_accel_target_wide.cpp
@@ -716,9 +716,9 @@ static void process_dma_readbuf()
 
     if (g_scsi_dma.wide)
     {
-        // Read 16 bits per IO word. Per-byte parity is checked on each word
-        // and failures are OR-accumulated so that errors never cancel.
-        uint32_t parity_errors = 0;
+        // Read 16 bits per IO word. Parity is verified on each word so that
+        // errors cannot cancel across words.
+        bool parity_error = false;
         while (src + 4 < end)
         {
             for (int unroll = 0; unroll < 4; unroll++)
@@ -726,11 +726,7 @@ static void process_dma_readbuf()
                 uint32_t word = *src++;
                 *(uint16_t*)dst = (uint16_t)word;
                 dst += 2;
-                uint32_t w2 = (word >> 4) ^ word;
-                uint32_t w3 = (w2 >> 2) ^ w2;
-                uint32_t w4 = (w3 >> 1) ^ w3;
-                uint32_t p2 = ((uint64_t)(word & 0x30000) * 0x810000) >> 32;
-                parity_errors |= (~(w4 ^ p2)) & 0x0101;
+                if (!scsi_check_parity_16bit(word)) parity_error = true;
             }
         }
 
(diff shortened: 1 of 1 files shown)


lib/ZuluSCSI_platform_RP2MCU/scsi_accel_target_wide.cpp
```diff
@@ -716,16 +716,21 @@ static void process_dma_readbuf()
...
             {
                 uint32_t word = *src++;
                 *(uint16_t*)dst = (uint16_t)word;
                 dst += 2;
-                parity ^= word;
+                uint32_t w2 = (word >> 4) ^ word;
+                uint32_t w3 = (w2 >> 2) ^ w2;
+                uint32_t w4 = (w3 >> 1) ^ w3;
+                uint32_t p2 = ((uint64_t)(word & 0x30000) * 0x810000) >> 32;
+                parity_errors |= (~(w4 ^ p2)) & 0x0101;
```
PetteriAimonen: I'd like this code to remain in the `ZuluSCSI_platform_gpio_Wide.h`, just in case some future hardware version wants to use different GPIO mapping.

Maybe just use a `bool parity_errors` and `scsi_check_parity_16bit()`? It might optimize to about equal amount of instructions.

PetteriAimonen (changes requested):
I think it would be better to use the parity test inline functions from the GPIO header, in order to make future maintenance easier.

I agree that the performance impact is probably negligible, but that should be verified by testing at the highest throughput scenario (20 MHz sync bus in both narrow and wide configuration).

PetteriAimonen (approved):
Code looks good to merge.

Performance effects on write speed in SYNC-20 mode are worth checking, but I think that this won't cause any noticeable performance degradation.

Issue #846, opened 2026-05-08
Tape Emulation - can't read variable block tapes that were just written


jgeorge44:
This is a new setup so it may be as much user error as anything else.

Hardware: ZuluSCSI  Compact RP2040
Firmware:  2026.04.15-release Apr 15 2026 21:28:33
Host: generic intel PC, Ubuntu 20.04.06 LTS, kernel 5.4.0-216-generic #236  

I have set up the board to emulate a Tandberg TDC3820 QIC-525MB drive. sginfo returns:

```$ sudo sginfo /dev/sg1
INQUIRY response (cmd: 0x12)
----------------------------
Device Type                        1
Vendor:                    TANDBERG
Product:                   TDC3820
Revision level:            2.0
```
...which matches my zuluscsi.ini configuration.

I have a tape image, created with Eric Smith (brouhaha)'s tapeutils. It's an OS/400 PTF (patch) tape, It's an IBM Standard Label (SL) tape written with multiple files and variable block sizes. It's attached as ptftape.tap.

I can write this .tap file to the zuluscsi emulated tape drive with "tapecopy ptftape.tap /dev/st1" successfully.

However, if I try to read the image back from the emulator, it fails almost instantly when trying to read the tape with a block size mismatch. Logs attached.

To Repro:
- set up zuluscsi with the attached ini file. 
- create an empty file SCRATCH_1.tap on the SD card (touch SCRATCH_1.tap)
- write the attached ptftape image to the emulator with "tapecopy ptftape.tap /dev/stX" (X=whatever your drive is)
- note lack of any errors (see zuluwrite.log)
- read the emulated tape back with "tapecopy /dev/stX ptftest.tap"
- note almost immediate failure (see zuluread.log)
- log attached, but notably the error is pretty clear:
"------ TAP block length mismatch: block size=32768 bytes read=39602"

As far as I can tell there is no block larger than 32K written to the tape, so there should be no 39K read happening anywhere.

Attached also is the SCRATCH_1.tap file from the SD card, containing the tape that the emulator wrote. 

[zuluread.log](https://github.com/user-attachments/files/27541371/zuluread.log)
[zuluwrite.log](https://github.com/user-attachments/files/27541372/zuluwrite.log)
[zuluscsi.ini.txt](https://github.com/user-attachments/files/27541380/zuluscsi.ini.txt)
[zulureadorig.txt](https://github.com/user-attachments/files/27541481/zulureadorig.txt)

[...]

PR #848, opened 2026-05-09
Fix variable tape reads with tap file format

morio:
User reported issue with reading a tape image that was just created here: https://github.com/ZuluSCSI/ZuluSCSI-firmware/issues/846

Reading variable SIM-H tap tape was trying to fill the cache with multiple reads. This fix fills the cache with a single variable block read then writes it out over the SCSI interface.

There still seems to be some sort of issue with reading and writing using https://github.com/brouhaha/tapeutils/blob/master/tapecopy.c The resultant copy from ZuluSCSI is about twice the size of original image copied to the ZuluSCSI.


jgeorge44:
With this last build `2026.05.15-dev May 15 2026 22:20:48` I am able to write an aws tape to completion without SDIO errors and I am also able to read that image back with hercules tapecopy to completion without errors. woohoo!

Issue #900, opened 2026-07-06
Daynaport SCSI Ethernet emulation breaks after 1000 seconds (ZS pico slim)


vom513:
Just recently started playing with this.  I have it working on my Mac SE 30.  Firmware is 2026-07-01.

In both MacOS 7.5.5 (OpenTransport 1.3.1 (?)) and NetBSD 11.0_RC5 (dse driver) I have the same behavior.  When the system first comes up network is useable.  After a bit (maybe 30 minutes ?) I get no network traffic (no errors).  

Log and config attached.

[zuluscsi.ini.txt](https://github.com/user-attachments/files/29716796/zuluscsi.ini.txt)
[zululog.txt](https://github.com/user-attachments/files/29716797/zululog.txt)

aperezbios:
@vom513 thanks for the report. When you say it "breaks", does the disk emulation continue working as expected, or does it also stop working? Also,which ZuluSCSI firmware version were you using prior to this upgrade?

vom513:
Disk keeps on working fine - no errors or issues.

As for firmware versions - I just got this Mac SE30 in the last month - so I've only had a few recent (June/July 2026) firmwares in play for it.  This is the first time I have ever used the wifi on it.

Also, I'm letting the Mac run (in System 7) for a bit to see if it falls of the wifi AP.  These are Aruba 515s - decent enterprise grade.  I have lots of IOT / Pis etc along with the usual devices on them with no issues.  Unless there is a very specific bad interaction between the ZSPS wifi chip and these APs - I don't think it's the network.

vom513:
Okay I might have found something.  I had said "maybe 30 minutes".  I think it's more like 20 min (a little less actually).

In general - especially on the Mac system 7 side - I would use Fetch (FTP client) to pull over a piece of software to play with.  Play with this - and then restart or even shut down.  It's just today that I finally let it sit for a while and then tried to use the network.

The only thing I have "automatically" active on the Mac SE 30 is an NTP client that only runs on startup.  So after that it's completely quiet.  The Aruba software (Instant AP flavor, 8.10.x.x version) has a 1000 second client timeout (I think most Arubas have the same 1000 second timeout regardless of firmware / management flavor).

Check this out:

```
[...]

PR #901, opened 2026-07-10
Send keep alive packet to avoid silent deauth

morio:
With some routers the WiFi connection would deauthenticate and the cyw32 driver would not detect it. This fix sends a dummy packet every 120 seconds, which can be changed a value in seconds or disabled by setting `WiFiKeepAliveSecs` to a value or 0 respectively under the `[SCSI]` section in `zuluscsi.ini`.

The solution is from @vom513 published in issue: https://github.com/ZuluSCSI/ZuluSCSI-firmware/issues/900 as a patch.


aperezbios:
As per @vom513's comment in #900, this has worked for more than 24 hours.

PR #902, opened 2026-07-15
Enable firmware upgrade of Zulu Control board via i2c (zulucontrol.uf2)

morio:
This allows a board running ZuluControl-firmware to be updated via I2C by sending over the `zulucontrol.uf2` file on the root of the SD card. On a successful update the file is deleted.


aperezbios:
Testing cycle completed, merging.

PR #916, opened 2026-08-19
Fix AS/400 Skip Read/Write 512-byte assumptions causing install-time data corruption

PoC-dev:
## Problem

On `PLATFORM_AS400` with non-512-byte sectors (520/522 bytes), OS/400 installs to
an emulated DASD would appear to format and install successfully, then crash with
SRC A6004403 when switching from limited paging to full paging. Root cause is a
cluster of places in `diskDataOut()`/`scsiDiskStartRead()` (`src/ZuluSCSI_disk.cpp`)
that implicitly assumed a 512-byte-aligned world, none of which hold for AS/400
sector sizes:

1. **Skip Write truncation**: the SD write pipeline chunks data in sizes derived
   from buffer availability, not sector size. Skip Write (`0xEA`) walked its chunk
   sector-by-sector but truncated to `len - (len % bytesPerSector)` while the byte
   counter driving the pipeline still advanced by the full (unaligned) `len` —
   silently dropping up to `bytesPerSector - 1` bytes at chunk boundaries that
   didn't land on a sector boundary (nearly all of them, since chunk sizes are
   platform/buffer-driven, not sector-driven).
2. **Skip Write mask position lost across multi-chunk transfers**: `diskDataOut()`
   caps how much it processes per invocation to what fits in one SD write buffer
   (~125 sectors at 522 bytes/sector with the default 64KB buffer) and re-enters
   for the remainder when a command exceeds that. AS/400 Skip commands allow up to
   256 blocks per the IBM spec, so any mask over ~125 sectors — entirely legal —
   needs a second invocation. `g_disk_transfer.skip_command` was being cleared
   unconditionally at the end of *every* invocation rather than only when the
   whole command finished, so the second invocation silently fell through to a
   plain contiguous write for the remainder, ignoring the mask entirely.
3. **Skip Read bypassed by the prefetch cache**: the read-ahead prefetch cache
   holds sectors from a prior ordinary contiguous read and was consulted
   unconditionally before deciding how to serve a read. A Skip Read gathers
   sectors per its mask, which can skip/reorder relative to a straight LBA run, so
   serving it from the cache would silently return contiguous data instead.
   Latent — inert whenever `PrefetchBytes=0`, which is why it wasn't observed
   directly, but would corrupt reads the moment prefetching is enabled.

None of this is specific to one sector size — everything here is driven by the
runtime `bytesPerSector`, not a hardcoded constant, so it applies equally to 520-
and 522-byte AS/400 sectors.

Also included: a one-line `platformio.ini` restore. An unrelated upstream merge
(`7422a76`, "Merge branch 'LTS/V1.x' into main") silently dropped the
`ZuluSCSI_RP2MCU`/`_Wide`/`_RP2040`/`_Pico`/`_Blaster`/etc. build environments,
which made this branch (forked from `main`) unable to build `ZuluSCSI_Wide` at
all. Folded in here so the branch is independently buildable; happy to split it
out if maintainers would rather review it separately.

## Fix

- Round the SD-write chunk length down to a whole sector *before* it's consumed
  for anything (not after), so the byte-accounting, the actual write, and the
  SCSI-side "finished" bookkeeping all agree on the same already-aligned amount.
  The unaligned remainder stays in the SCSI buffer and is picked up whole once the
  next chunk completes it.
- Only clear `skip_command` (and the associated mask/position state) once a Skip
  command has actually finished, failed, or the bus was reset — not after every
  `diskDataOut()` invocation.
- Skip the prefetch-cache lookup entirely when a Skip Read is in progress.

## Testing

Built and tested on real hardware (ZuluSCSI Wide, IBM 9401-150 running OS/400
V4R4). Prior to this fix, install reliably failed at the full-paging transition
with SRC A6004403 after correct disk sizing was in place. With all of the above
applied, a full OS/400 install (format → SLIC install → library restore →
environment switch) completed end-to-end with no errors, no SCSI CHECK_CONDITIONs
on any Read10/Write10/Skip Read/Skip Write, and no truncated/mask-exhausted Skip
transfers, across ~216 minutes and 9,000+ Skip commands.

### DIFF
--- a/src/ZuluSCSI_disk.cpp
+++ b/src/ZuluSCSI_disk.cpp
@@ -3130,17 +3130,7 @@ void scsiDiskStartRead(uint32_t lba, uint32_t blocks)
 
 #ifdef PREFETCH_BUFFER_SIZE
         uint32_t prefetch_sectors = 0;
-        const uint8_t *prefetch_ptr = NULL;
-#ifdef PLATFORM_AS400
-        // The prefetch cache holds sectors from a prior ordinary contiguous
-        // read. An AS/400 Skip Read (the linked Read10 that follows a Skip
-        // Read mask CDB, dispatched here with skip_command already set to
-        // 0xE8) must gather its data per the skip mask instead - serving it
-        // from the cache would return contiguous data and silently bypass
-        // the mask.
-        if (g_disk_transfer.skip_command == 0)
-#endif
-        prefetch_ptr = scsiDiskPrefetchRead(img.scsiId, transfer.lba, bytesPerSector, &prefetch_sectors);
+        const uint8_t *prefetch_ptr = scsiDiskPrefetchRead(img.scsiId, transfer.lba, bytesPerSector, &prefetch_sectors);
 
         if (prefetch_ptr)
         {
@@ -3604,8 +3594,19 @@ void scsiDiskSkip(uint32_t lba, uint32_t blocks, uint8_t mask_length,uint8_t ski
     }
     else
     {
-        g_disk_transfer.skip_command = skip_command;        
+        g_disk_transfer.skip_command = skip_command;
         g_disk_transfer.skip_position = 0;
+
+        if (skip_command == 0xE8)
+        {
+            // A Skip Read must gather its data per the mask; invalidate any
+            // sectors already sitting in the prefetch cache from a prior
+            // ordinary contiguous read now, so the linked Read10 that
+            // follows can't be served contiguous data that would silently
+            // bypass the mask. No-op if nothing is cached for this ID, or
+            // if PREFETCH_BUFFER_SIZE isn't defined at all.
+            scsiDiskPrefetchInvalidate(scsiDev.target->targetId);
+        }
 
         // Support optional linked command (CDB byte 9 bit 0)
         if (scsiDev.cdb[9] & 1)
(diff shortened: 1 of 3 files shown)


aperezbios:
@PoC-dev awesome!

This PR unintentionally changes the version string in main to the V1.x LTS string, which is not correct. Please fix and amend the PR. We need you to fix this trivial issue before we can accept the pull request.

PoC-dev:
Whoops. I'm admittedly not really fluent with the build system quirks. Should be correct now.

src/ZuluSCSI_disk.cpp
```diff
@@ -3083,7 +3130,17 @@ void scsiDiskStartRead(uint32_t lba, uint32_t blocks)
...
-        const uint8_t *prefetch_ptr = scsiDiskPrefetchRead(img.scsiId, transfer.lba, bytesPerSector, &prefetch_sectors);
+        const uint8_t *prefetch_ptr = NULL;
+#ifdef PLATFORM_AS400
+        // The prefetch cache holds sectors from a prior ordinary contiguous
+        // read. An AS/400 Skip Read (the linked Read10 that follows a Skip
+        // Read mask CDB, dispatched here with skip_command already set to
+        // 0xE8) must gather its data per the skip mask instead - serving it
+        // from the cache would return contiguous data and silently bypass
+        // the mask.
+        if (g_disk_transfer.skip_command == 0)
```
PetteriAimonen: The dangling `if` may accidentally break if code outside the `#ifdef` would later be modified.

Maybe you could call `scsiDiskPrefetchInvalidate()` when a skip command starts?

PetteriAimonen (approved):
Just a small suggestion, otherwise looks ok.

aperezbios (approved):

PR #918, opened 2026-08-21
Warnings Cleanup

erikarn:
Here's (i hope!) a set of warnings cleanup that doesn't break any functionality.

Most are straight up warnings cleanups (like typecasting where required for printf, etc) but a couple of things (like assembling file paths) were potentially going to overflow a buffer and in that case I figured erroring out was better than partially assembling a path.

For now it's been compile tested. I'll test on a couple of physical devices shortly.

### DIFF
--- a/lib/ZuluSCSI_Control/ZuluSCSI_control_api.cpp
+++ b/lib/ZuluSCSI_Control/ZuluSCSI_control_api.cpp
@@ -287,6 +287,7 @@ int controlListImages(uint8_t scsi_id,
     FsFile entry;
     while (entry.openNext(&dir, O_RDONLY))
     {
+	int ret = 0;
         if (!entry.getName(name, sizeof(name)) || entry.isHidden())
         {
             entry.close();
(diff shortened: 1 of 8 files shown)


lib/ZuluSCSI_Control/ZuluSCSI_control_api.cpp
```diff
@@ -301,9 +302,18 @@ int controlListImages(uint8_t scsi_id,
             continue;
 
         if (root_dir)
-            snprintf(full_path, sizeof(full_path), "/%s", name);
-        else
-            snprintf(full_path, sizeof(full_path), "%s/%s", imgdir, name);
+            ret = snprintf(full_path, sizeof(full_path), "/%s", name);
+	else
+            ret = snprintf(full_path, sizeof(full_path), "%s/%s", imgdir, name);
```
PetteriAimonen, line 307: There seems to be some indentation error (tab vs. spaces) here, and above on line 290.

PetteriAimonen (approved):
Looks good, except for some small tab vs. space indent issue.

lib/ZuluSCSI_Control/ZuluSCSI_control_api.cpp
aperezbios: @erikarn this can be fixed in a follow-on PR. No need for this to gate the rest.

PR #925, opened 2026-08-23
Add AS/400 disk profile loader with multi-DASD support

PoC-dev:
Lets a [SCSIn] section load a real captured AS/400 drive identity
(INQUIRY/VPD/MODE SENSE, from as400_disk_definitions.txt via the extractor
script) instead of always presenting the single built-in profile, and
extends that to multiple simultaneous SCSI IDs each presenting a different
real DASD unit.

- AS400_DiskProfile=<name> in [SCSIn] loads a named profile's identity for
  that SCSI ID. as400_disk_definitions.txt ships with nine real captured
  profiles (520/522-byte, CISC/PPC era) plus one verified-byte-perfect
  fabricated entry matching the built-in identity - usable out of the box,
  no AS/400 DASD hardware required. utils/extract_as400_disk_data.sh adds
  more drives when you have real hardware to capture from.
- Missing image files for a profiled ID are auto-created at the profile's
  real captured size.

Verified on real AS/400 hardware (9401-150): single-ID identity loading,
auto-image-creation, and three simultaneously profiled SCSI IDs all showing
correct, distinct identities in Dedicated Service Tools with no data
corruption. Actually using the presented disks has not yet been exercised.

Known limitation, documented in the README: using the *same* profile for
more than one SCSI ID currently gives them identical serial numbers, since
loaded-profile data is copied verbatim with no per-ID override yet. Use a
different profile per ID for now.

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,7 +50,15 @@ For AS/400 `AS400_DiskProfile=` SCSI IDs (see below), a correctly-sized image is
 
 AS/400 disk profiles
 ---------------------
-For AS/400 (`System = "AS400_PPC"` or `"AS400_CISC"`), multiple SCSI IDs can each emulate a different real DASD unit by setting `AS400_DiskProfile = "<name>"` in that ID's `[SCSIn]` section, naming a section captured by `utils/extract_as400_disk_data.sh` into `as400_disk_definitions.txt` on the SD card.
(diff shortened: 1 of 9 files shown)


morio (approved):
Looks good.

Maybe a later change would be to split the disk definitions into their own ini files with the model number in the filename and stick them in a sub directory so the ini parser would do less work.

PR #933, opened 2026-08-27
AS/400 CISC/PPC tape support, subdirectory custom-inquiry fix, and Blaster memory fixes

PoC-dev:
## Summary
- Add AS/400 tape identity support for CISC-era ("IBM 4100") and PPC-era ("IBM SLR5") tape drives, captured from real hardware -- INQUIRY, MODE SENSE, and VPD wiring, selected via `Device = "AS400_CISC"`/`"AS400_PPC"` same as disk.
- Fix: `parseCustomInquiryData()` (the entry point for all AS/400 custom identity/VPD/serial data, and generic `vpdXX=`/`spd=` INI overrides) was only ever called from the root-directory image scan, never from the per-ID default-subdirectory path (`TP0/`, `HD0/`, etc.) -- any device configured the conventional subdirectory way silently got a generic identity instead. Confirmed root-caused and fixed; verified end-to-end on real hardware via a temporary diagnostic (both storage and SCSI-bus-level serving confirmed correct, then diagnostic reverted).
- Add LOG SENSE page 0x02 (Write Error Counter) support for AS/400 tape -- real AS/400 hosts query this routinely regardless of per-ID configuration, confirmed via a real boot log showing repeated rejections during Main Storage Dump.
- Add `ZuluSCSI_Blaster_AS400`, a minimal build environment for real AS/400 use on Blaster (drops networking/DaynaPORT/audio/UI, none of which OS/400 has drivers for), plus the RAM-overflow fix needed once tape support was combined with the existing disk-profile work on this board.

## Test plan
- [x] Build succeeds
- [x] Hardware-tested across both ZuluSCSI Wide and Blaster, against real AS/400 hardware (9401-150): PPC tape identity both load and serve correctly, confirmed via a temporary diagnostic trace (storage: all VPD pages land; lookup: every real captured page is a clean HIT)
- [ ] Hardware test of the `ZuluSCSI_Blaster_AS400` build itself -- not being done on real AS/400 hardware, because of follow-up issues yet to be diagnosed.


PoC-dev:
I have tried the whole afternoon to investigate why RAM usage on the Blaster jumped more than it should have, but only when using the the CI pipeline. Here's a summary. I'm genuinely lost how to continue from here.

## ZuluSCSI_Blaster RAM overflow — investigation summary

**Issue**: This PR fails CI on `ZuluSCSI_Blaster` only, at link time (`cannot move
location counter backwards`, i.e. `.bss`+`.data` exceeds the 512KB RAM region).
All other environments (RP2040, Pico, Pico_2, Wide, DaynaPORT variants) build fine.

**Steps taken**:
- Reduced `MAX_CUSTOM_VPD_ENTRIES` for this env from 8 to 2 (saves 1548 bytes) — still overflows.
- Confirmed deterministic: re-ran the identical commit via an empty commit, same 1824-byte overflow both times.
- Confirmed environment parity: GCC 14.3.0 / binutils 2.43.1, and every pinned library/framework SHA, identical between a local build (Debian 12, succeeds with ~7KB RAM margin) and CI (Ubuntu 24.04, fails) — ruled out toolchain/package version mismatch.
- Confirmed unmodified upstream `main` (current tip, `MAX_CUSTOM_VPD_ENTRIES=8`) builds `ZuluSCSI_Blaster` successfully on CI today.
- Tested this PR's branch alone (no merge with `main`) vs. merged with `main`: branch alone overflows by 792 bytes; merged with `main` overflows by 1824 bytes — merging in `main`'s newer commits alone adds ~1032 bytes.
- Compared full per-object `.bss`+`.data` totals for every compiled object: this PR's own code totals **less** than baseline's (147176 vs 148724 bytes) — our own compiled code is smaller than what already ships, not bigger.
- Compared linker `--print-gc-sections` output (single-environment build, to avoid `-j8` interleaving) between this PR's failing build and baseline's passing build: the only `.bss`/`.data`-relevant difference found is one small networking-related variable (`g_wifi_join_secured`) newly kept live by this PR. Every other difference is DWARF debug metadata or flash-resident code/constants — irrelevant to a RAM overflow.

**Conclusion**: The overflow is real, deterministic, and reproducible on CI, but its exact source could not be isolated to a specific line of new code. Our own compiled objects measure smaller than baseline's; the one measurable RAM-relevant difference found (~1 byte) can't account for the ~800–1800 byte gap. The remainder likely comes from which prebuilt framework archive members (`libpico.a`/`liblwip.a`) get linked in due to symbol reachability — a difference invisible to per-object size comparison and not further isolated. Notably, this discrepancy **never reproduces locally**, even with identical toolchain/package versions confirmed byte-for-byte — only CI hits it.

**Open question for maintainers**: is `ZuluSCSI_Blaster`'s RAM margin already known to be this thin independent of this PR, and is trimming an existing feature (e.g. networking/DaynaPORT, matching what the newly-added `ZuluSCSI_Blaster_AS400` env already does) the expected path here, or is there a preferred way to pin down archive-level linking behavior on this board?

aperezbios:
Merging to branch to address GitHub runner CI build failure, before merging into main

PR #934, opened 2026-08-27
AS/400 CISC/PPC tape support, subdirectory custom-inquiry fix, and Blaster memory fixes

morio:
## Summary
- Add AS/400 tape identity support for CISC-era ("IBM 4100") and PPC-era ("IBM SLR5") tape drives, captured from real hardware -- INQUIRY, MODE SENSE, and VPD wiring, selected via `Device = "AS400_CISC"`/`"AS400_PPC"` same as disk.
- Fix: `parseCustomInquiryData()` (the entry point for all AS/400 custom identity/VPD/serial data, and generic `vpdXX=`/`spd=` INI overrides) was only ever called from the root-directory image scan, never from the per-ID default-subdirectory path (`TP0/`, `HD0/`, etc.) -- any device configured the conventional subdirectory way silently got a generic identity instead. Confirmed root-caused and fixed; verified end-to-end on real hardware via a temporary diagnostic (both storage and SCSI-bus-level serving confirmed correct, then diagnostic reverted).
- Add LOG SENSE page 0x02 (Write Error Counter) support for AS/400 tape -- real AS/400 hosts query this routinely regardless of per-ID configuration, confirmed via a real boot log showing repeated rejections during Main Storage Dump.
- Fixed Blaster memory issue form original pull request: https://github.com/ZuluSCSI/ZuluSCSI-firmware/pull/933
## Test plan
- [x] Build succeeds
- [x] Hardware-tested across both ZuluSCSI Wide and Blaster, against real AS/400 hardware (9401-150): PPC tape identity both load and serve correctly, confirmed via a temporary diagnostic trace (storage: all VPD pages land; lookup: every real captured page is a clean HIT)


aperezbios (approved):
Basic functionality testing completed based on CI builds.

PR #936, opened 2026-08-31
Fix permanent SD-card log lockout after opening the media menu's image list

PoC-dev:
show_image_list() took log_lock() to guard its screen output but never released it - none of the three ways to leave MEDIA_MENU_IMAGE_LIST (cancel, invalid index, successful load) called log_unlock() either. save_logfile() checks the lock and bails out immediately if held, so opening this screen once silently stopped all SD-card log writes for the rest of that boot, with console output unaffected.

Fixed at the source: release the lock at the end of show_image_list() itself, right after the screen finishes printing, matching how every other menu action in this file already brackets log_lock()/log_unlock() tightly around the print instead of holding it across a wait for input.


aperezbios:
Good catch, thanks for the bugfix!

PR #938, opened 2026-09-01
Fix Test Unit Ready for removable devices with no SD

morio:
The TEST UNIT READY command was retuning a check condition for hard drives that hadn't started or there was no image when the device was actually removable. This fix returns Medium Not Present for these devices when on image is present, most likely because the SD card was removed.

This is attempting to address an emulated MO device not reading reading media after the SD card was removed and reinserted.


aperezbios:
Customer confirms this fixes their issue.

PR #947, opened 2026-09-05
Add preset for IRIX system

morio:
A user on a SGI Tezro reported a certain timing combination worked at TurboMax (251.2MHz). Implemented a system preset that applies those timings:

```
g_zuluscsi_timings->scsi_20.rdelay1 = 9;
g_zuluscsi_timings->scsi_20.rtotal_period_adjust = 3;
```

Named the System Setting `IRIX`.
```
[SCSI]
System = "IRIX"
```
Should apply the settings


techomancer:
Didnt work. Not sure if because of timings but early in kernel boot I was getting a lot of
ql0d1: HARD ERROR: completion status [0x14] (SCS_UNEXPECTED_BUS_FREE), scsi status[0x0], state flags[0x2f00], status flags[0x80]

PR #950, opened 2026-09-08
Bake the exact git commit + build timestamp into the firmware reliably

PoC-dev:
## Summary
- Adds `src/inject_build_info.py`, a PlatformIO pre-build script that runs on every build invocation and writes a small generated header (`src/ZuluSCSI_build_info_generated.h`, gitignored) containing the exact git commit (+ a dirty-tree flag) and a real build timestamp.
- `ZuluSCSI_log.cpp` includes this header via `__has_include` and uses it for the boot log's `FW Version` line, falling back to `__DATE__`/`__TIME__` and `"unknown"` for build environments that don't wire the script in.
- Motivation: `__DATE__`/`__TIME__` only refresh when the compiler actually recompiles that specific translation unit. An object-level build cache (e.g. ccache with `time_macros` sloppiness) can silently serve a stale object for that one file even after touching it, with the only symptom being a wrong version string on the device.
- Only `ZuluSCSI_log.cpp` includes the generated header, so a differing build forces a recompile of that one file + relink, not a full project rebuild.
- Wired into `[env:ZuluSCSI_RP2MCU]`'s `extra_scripts` (covers Blaster/Wide/RP2040/Pico); the GD32 V1.x line is untouched.

## Test plan
- [x] Verified `strings firmware.bin | grep git:` reports the correct commit hash and a fresh ISO-8601 timestamp after a normal `pio run` build, with no rebuild/clean step required.
- [x] Confirmed via a real regression this behavior was designed to catch: an earlier version of this same fix (compiler-define based instead of header based) silently never took effect at all.


morio (approved):
Looks good, thanks!

PR #951, opened 2026-09-08
scsi: SCSI-2 compliance fixes for target disconnect/reselection

PoC-dev:
## Summary
Five standards-compliance corrections to the target disconnect/reselection primitive layer, on top of this branch's own target disconnect/reselection implementation:
- `42e0c14` -- fix `scsiPhyReselect()` arbitration and SEL timing per SCSI-2
- `0b0181f` -- enforce the disconnection delay before reselection
- `8374747` -- service MESSAGE OUT if ATN is asserted at reconnection
- `1740c43` -- reject selection with more than two ID bits asserted
- `3d67f3c` -- honor the negotiated disconnect time limit (mode page 0x02)

This is universal, platform-wide SCSI PHY/timing code (not AS/400-specific -- no `#ifdef PLATFORM_AS400` anywhere in this branch) affecting every real backend and the whole device/host matrix, not just AS/400 hardware. Please test broadly, not only against AS/400 configurations.

> **Note:** This does **not** implement Disconnect/Reconnect per se which is entirely optional per spec. Because the Zulu boards lack the delay of mechanical components, relinquishing the bus to be nice to other devices might not justify touching core components of the basic SCSI layer. But the basic foundation for a later expansion will be more sound with this addition.

## Test plan
- [ ] Needs testing across ZuluSCSI's device/host matrix generally -- this session's own testing was AS/400-focused (see the companion AS/400 CISC fixes PR) and doesn't cover other host families.


PoC-dev:
Agreed, completely botched. Will do fresh.

PR #952, opened 2026-09-08
as400: fix CISC (9401-P02) IPL failures -- unit attention, disk state, WRITE BUFFER, VPD data

PoC-dev:
## Summary
Fixes a multi-layered set of bugs that together prevented a 9401-P02 (CISC) AS/400 from ever completing IPL against a ZuluSCSI-emulated DASD. Root-caused via extensive real-hardware Ancot SCSI analyzer traces and live console logging against actual AS/400 hardware. With all of these fixed, a real P02 now completes IPL through the OS/400 main menu.

- `c46ff30` -- fix another stack overflow in the AS/400 disk-profile loader (`parseCustomInquiryData()`'s own stack-local buffer, missed by an earlier fix to the same bug shape one frame further in)
- `5d07511`/`87dc5cd` -- don't suppress unit attention for hosts that select without ATN but have explicitly asked for it (`EnableUnitAttention=Yes`, set by both the AS/400 and DOS presets). The old blanket "disable UA for any ATN-less selection" rule (there to protect genuinely broken SCSI-1 hosts like the Mac Plus) was wiping the P02 SP's pending power-on unit attention before its first TEST UNIT READY could ever see it -- confirmed byte-for-byte against a real disk's own Ancot baseline.
- `91ea5b1`/`4952605`/`08f6d47` -- `doTestUnitReady()` gates its entire CHECK_CONDITION/NOT_READY sequence on a `started` flag that `scsiReset()` never cleared on a bus reset (only `scsiInit()` did, once, at boot) -- independent of unit attention entirely. A P02 power-cycle, or even a routine mid-session bus reset, left every subsequent TEST UNIT READY returning GOOD immediately with no fresh unit attention ever shown. Both fixes generalized beyond AS/400+FIXED-disk-only to any device type under a quirk that wants unit attention.
- `d3fb275` -- fix `WRITE BUFFER`'s mode field being read from the wrong buffer (`scsiDev.data[1]`, stale payload data from the previous command, instead of `scsiDev.cdb[1]`). Universal bug, not AS/400-specific. Found because OS/400's own LIC-reload IPL step uses `WRITE BUFFER` mode 5 (download microcode and save); acceptance was effectively random depending on stale buffer content. Also widened accepted modes to include download-microcode (4/5) alongside the existing loopback/speed-test modes (0/2) -- accepted and discarded, since Zulu has no real on-disk microcode to update.
- `942215a` -- corrected a stale byte in the `45G9463`/`45G9463-1` disk profiles' VPD page 0x03 (a firmware/microcode revision field), verified against a fresh live capture from both real physical disks. This was introduced by an unrelated *downgrade* of the P02 from V3R2 to V2R3. **Note:** Handling this properly will be part of another PR in the future.

## Test plan
- [x] Real 9401-P02 (CISC) hardware IPL confirmed reaching the OS/400 main menu, with I/O activity ceasing normally (all IPL steps complete).
- [x] Confirmed through both a mid-session bus reset and a full P02 power-cycle without rebooting Zulu itself.
- [ ] Not yet confirmed on a PPC-generation AS/400 (structurally covered by the same code path, but this session's hardware access was CISC-only).


morio (approved):
Thanks for the bug fixes

aperezbios:
@PoC-dev this is confirmed working on a RISC AS/400, model 9404-600, using the resulting CI build, booting as DASD. As such, I've accepted the PR. Thanks again.

aperezbios:
@PoC-dev a portion of this PR has been confirmed to break the ability to boot from CD-ROM on our AS/400 Model 600 RISC machine. We have confirmed that reverting a portion of the changes (just scsi.c) in this commit resolves ability to boot on RISC, for us:

https://github.com/ZuluSCSI/ZuluSCSI-firmware/commit/5c1de1449a070bc6d9c0fabb9cd8e0faa805762f

If you could incorporate the commit above, which is in the `revert/back-out-changes-breaking-as400-cdrom` branch, into a new PR that resolves this issue on AS/400 RISC machines, that would be greatly appreciated.

PoC-dev:
Your  fix is too broad, it also undos the unit attention dance for tape drives being required on the 9401-P02. Will narrow this down to CD drives only. CISC machines don't support CD anyway.

PR #953, opened 2026-09-08
scsi: SCSI-2 compliance fixes for target disconnect/reselection

PoC-dev:
## Summary
Five standards-compliance corrections to the target disconnect/reselection primitive layer, including the foundational implementation itself (pulled in from @morio's `feature/enable-target-disconnect-reselection`, since this branch builds directly on it):
- `842ed48` -- Implement target disconnect and reselection (morio's original commit, carried forward)
- `ec5f454` -- fix `scsiPhyReselect()` arbitration and SEL timing per SCSI-2
- `d23dddd` -- enforce the disconnection delay before reselection
- `ad741aa` -- service MESSAGE OUT if ATN is asserted at reconnection
- `702e434` -- reject selection with more than two ID bits asserted
- `0c9adf6` -- honor the negotiated disconnect time limit (mode page 0x02)

This is universal, platform-wide SCSI PHY/timing code (no `#ifdef PLATFORM_AS400` anywhere) affecting every backend and device/host matrix, not just AS/400 hardware. Please test broadly.

> This does not implement full Disconnect/Reconnect policy (entirely optional per spec) -- it's the standards-compliance foundation for later use.

Supersedes #951, rebased cleanly onto current `main` (11 files, 634 insertions(+), 164 deletions(-), no unrelated diff noise).

> **Note:** This does not implement Disconnect/Reconnect per se, which is entirely optional per spec. Because the Zulu boards lack the delay of mechanical components, relinquishing the bus just briefly to be nice to other devices might not justify touching core components of the basic SCSI layer. But the basic foundation for a later expansion will be more sound with this addition.

## Test plan
- [x] Diff verified clean against current `main` tip
- [ ] Needs testing across the device/host matrix -- this session's testing was AS/400-focused


lib/SCSI2SD/src/firmware/scsi.c
```diff
@@ -1058,7 +1068,8 @@ static void process_SelectionPhase()
 		// Save our initiator now that we're no longer in a time-critical
 		// section.
 		// SCSI1/SASI initiators may not set their own ID.
-		scsiDev.initiatorId = (selStatus >> 3) & 0x7;
+		scsiDev.initiatorId =
+			(g_scsi_sts_selection_initiator <= 7) ? g_scsi_sts_selection_initiator : -1;
```
morio:
```suggestion
			(g_scsi_sts_selection_initiator < S2S_MAX_TARGETS) ? g_scsi_sts_selection_initiator : -1;
```

There are probably no initiators on the upper scsi IDs, but just to be safe

lib/ZuluSCSI_platform_RP2MCU/scsiPhy.cpp
```diff
@@ -57,34 +57,71 @@ extern "C" bool scsiStatusBSY()
...
 volatile uint8_t g_scsi_ctrl_bsy;
 
-void scsi_bsy_deassert_interrupt()
+static uint8_t pack_selection_status(uint32_t sel_bits)
 {
-    if (SCSI_IN(SEL) && !SCSI_IN(BSY))
+    // SCSI-2 6.1.3: a target shall not respond to a selection if more
+    // than two SCSI ID bits are asserted on the DATA BUS.
+    int bitCount = 0;
+    for (int id = 0; id < 8; id++)
```
morio:
```suggestion
    for (int id = 0; id < S2S_MAX_TARGETS; id++)
```
Wide support

lib/ZuluSCSI_platform_RP2MCU/scsiPhy.cpp
```diff
@@ -57,34 +57,71 @@ extern "C" bool scsiStatusBSY()
...
         }
+    }
+
+    if (sel_id < 0)
+    {
+        return 0;
+    }
+
+    g_scsi_sts_selection_initiator = 0xFF;
+    for (int id = 0; id < 8; id++)
```
morio:
```suggestion
    for (int id = 0; id < S2S_MAX_TARGETS; id++)
```

lib/ZuluSCSI_platform_RP2MCU/scsiPhy.cpp
```diff
@@ -127,6 +164,87 @@ extern "C" bool scsiStatusSEL()
     return SCSI_IN(SEL);
 }
 
+extern "C" bool scsiPhyReselect(uint8_t targetId, uint8_t initiatorId)
+{
+    if (initiatorId > 7 || targetId >= S2S_MAX_TARGETS)
```
morio:
```suggestion
    if (initiatorId >= S2S_MAX_TARGETS || targetId >= S2S_MAX_TARGETS)
```

morio (changes requested):
Though probably not found out in the wild, I don't think the SCSI spec prohibits initiators living on the upper Wide SCSI IDs.

Thanks for getting CISC AS400 working

PoC-dev:
You're correct to push back on those. Corrected.

PoC-dev:
Pushed f22ffff, which replaces the hardcoded 7/8 bounds with S2S_MAX_TARGETS at all four sites you flagged, and applies the same fix to the GD32F205/GD32F450 copies for consistency. Ready for another look.

morio (approved):
Thanks!

PR #954, opened 2026-09-09
as400: exclude CD-ROM from the EnableUnitAttention generalization (fixes SRC B1014507 regression on RISC)

PoC-dev:
## Summary
PR #952 widened the CISC-disk unit-attention/started-flag fix from `quirks==AS400 && deviceType==FIXED` to any device type gated on `EnableUnitAttention`. This regressed CD-ROM boot on RISC hardware: @aperezbios reported SRC B1014507 (generic CD-ROM read error) on a real AS/400 Model 600 doing a manual D-mode IPL from CD-ROM (#952, comment https://github.com/ZuluSCSI/ZuluSCSI-firmware/pull/952#issuecomment-5593635061). @morio confirmed a full revert of this file's #952 changes fixes it, on `revert/back-out-changes-breaking-as400-cdrom` (`5c1de14`).

This PR takes a narrower approach instead of that blanket revert: exclude `S2S_CFG_OPTICAL` specifically from all three generalized gates (`scsiReset()`, `process_SelectionPhase()`, `scsiInit()`), while keeping the fix for `S2S_CFG_FIXED` and `S2S_CFG_SEQUENTIAL` (disk + tape) intact -- both still needed and hardware-confirmed for AS/400 CISC (tested: P02, P03). Real optical drives auto-spin-up with no expected START UNIT handshake, unlike disk (and apparently tape); that's plausibly what this RISC boot ROM's minimal CD-boot path can't tolerate.

- `91f4b5f` -- exclude S2S_CFG_OPTICAL from the EnableUnitAttention generalization in scsiReset(), process_SelectionPhase(), and scsiInit()

## Test plan
- [x] Verify via emulated CD-ROM manual D-mode IPL against 9401-150 (not yet re-tested against the reporter's specific Model 600 -- would appreciate confirmation from @aperezbios/@morio)
- [x] Disk (original CISC fix) and tape (ongoing investigation) behavior unaffected -- only OPTICAL's gate changed


aperezbios:
@PoC-dev this fix is also confirmed to work on our 9404-600. I can successfully boot from CD-ROM.

PR #958, opened 2026-09-13
Add SD card partition passthrough (PARTITION=n)

PoC-dev:
## Summary
- Adds `PART=n` -- reference an SD-card partition by number instead of hand-computed `RAW:` sector ranges, resolved from the card's own MBR or GPT partition table via a new, self-contained reader (`src/ZuluSCSI_partition_table.h`/`.cpp`). This duplicates a small amount of SdFat's own internal (non-public) partition-table logic on purpose, to avoid patching the vendored library directly -- see the file header comment for the intended long-term path (upstreaming into SdFat). Worth contacting the vendor for a code change.
- Validates before ever presenting the SCSI ID: partition not found, too small for a configured `AS400_DiskProfile`'s required capacity, or overlapping another claimed partition are hard errors with specific log messages. Misalignment to the SD card's preferred AU_SIZE is a warning only, not a hard block -- real cards' AU_SIZE often exceeds the 1 MiB alignment convention most partitioning tools default to, so a hard block would reject correctly, conventionally-partitioned card.
- Tested end to end with a PC running Linux.


PoC-dev:
Changed the ini key according to @aperezbios request elsehwere.

morio (approved):
Code works
Switch config key from PART to Partition and added partition support for SCA configured SCSI IDs under the [SCSIn] section.

PR #961, opened 2026-09-16
Fix 6 Skip Read/Write and PrefetchBytes corruption bugs

PoC-dev:
## Summary
A targeted code review traced the Skip Read (0xE8) / Skip Write (0xEA) command paths and PrefetchBytes handling for possible data corruption. All six issues found trace back to pre-existing upstream commits (confirmed via `git log -L`), independent of and predating unrelated AS/400 work in progress on another branch.

1. `diskDataIn()`: `skip_command` was only cleared on a fully-successful Skip Read, leaving it stuck armed across a failed/interrupted multi-invocation Skip Read -- silently applied to the next ordinary Read10.
2. `scsiDiskReset()`: never cleared `skip_command`, despite clearing every other piece of in-flight transfer state -- a bus reset mid-Skip-Read/Write left skip state armed across the reset.
3. `scsiDiskStartRead()`: the linked Read10 that must follow a Skip Read was never validated against `skip_lba`/`skip_blocks`, unlike the write side's equivalent check.
4. `start_dataInTransfer()`: a Skip Read's own "fail instead of shipping stale bytes" guard didn't actually work -- the data-in callback ran unconditionally afterward and shipped the stale buffer tail before the host ever saw the CHECK_CONDITION.
5. `diskDataIn()`'s prefetch loop: `maxPrefetchBytes` is `int` (unclamped from `PrefetchBytes=`), compared against a `uint32_t` -- a negative value silently defeated the per-image prefetch clamp.
6. Same prefetch loop: a stale-looking guard set fields nothing in the file actually reads; only worked by coincidence. Fixed to set the field the callback actually checks.

## Test plan
- [x] Verified internally consistent (brace balance, `scsiIsWriteFinished()` semantics)
- [x] Hardware-verified via extended real-world AS/400 IPL/usage on this codebase's performance-optimization branch (which carries these same fixes) -- the two real-hardware failures seen during that testing were both independently root-caused to unrelated causes (stale pre-existing corrupted disk image data; a loose SCSI cable), not this code.


morio (approved):
Thanks for the fixes

PR #962, opened 2026-09-17
AS/400 write performance: AlignUnalignedAccesses, capacity-clamp fix, diagnostics

PoC-dev:
## Summary
- Adds `AlignUnalignedAccesses` (`off`/`cisc`/`ppc`/`auto`, per-device) -- pads/groups AS/400's 520/522-byte logical sectors onto the SD card's native 512-byte boundaries, trading SD space for reduced access latency. CISC pads each 520-byte sector to its own 1024-byte slot; RISC/PPC groups 8x522-byte sectors into a 4608-byte (9-SD-sector) physical group matching OS/400's own paging. Off by default; an existing un-gapped image keeps working unmodified after upgrading.
- Completes the `PART:n`/`Partition=n` capacity clamp (reported capacity now matches the bound `AS400_DiskProfile`'s own declared geometry) and fixes a write-chunk alignment bug (AS/400 writes now round to a whole `bytesPerSector` multiple unconditionally, not just for Skip Write).
- New tools: `utils/as400_gapconv` (converts an image between tightly-packed and gapped layouts) and `utils/as400_part_planner.py` (computes exact AU-aligned partition boundaries for `gdisk`, given a set of `as400_disk_definitions.txt` profiles to place).
- Fixes two `zululog.txt` bugs: an ordering bug where `LogToSDCard` was checked before it was actually loaded from ini, and a stale `g_rawdrive_active` leak that permanently blocked logging once any `PART:n`/`RAW:` config had been active.
- Adds temporary IOTrace instrumentation (`src/ZuluSCSI_iotrace.h`, gated by a new `IOTrace=` ini setting, off by default) for offline SD-access-pattern analysis during this investigation -- diagnostic only, no effect on disk-data integrity.
- Updates `README.md`/`README-as400.md` to match, including reconciling the `Partition=n` key naming with what already shipped on `main` via #958.

## Test plan
- [x] `AlignUnalignedAccesses` hardware-verified: extended real-world AS/400 IPL/usage on 9401-150 (RISC/PPC) and 9401-P02 (CISC), no issue traced back to it
- [x] Capacity-clamp and write-chunk-alignment fixes hardware-confirmed on real IPL
- [x] Merged current `main` in before opening this PR to keep the diff to exactly this branch's own content (no unrelated noise from #953/#958/#961)


PoC-dev:
I especially would appreciate feedback about the comprehensibility of `README-as400.md`.

morio (approved):
This is a long pull request. Thank you for the optimizations and iotrace will be very useful going forward.

PR #963, opened 2026-09-19
Direct support for Processor Device type

morio:
Fix new log messaging which erroneously stated that config files for SCSI network (DaynaPORT) or SCSI sound devices are not contiguous. For network and audio stream devices that are classified as SCSI processor devices, shortcut the ImageBackingStore initialization and directly issue the device is open and skip processing the filename.


morio:
Needs testing

aperezbios:
Tested and working as expected.

